Issue #1, opened 2018-10-03
Enable splunk as a Linux service


rcabral-splunk:
On splunk_common role, it would be helpful to decide whether to enable Splunk as a Linux service for those non-Dockerized installations. A feature toggle flag would be required to control whether to enable this installation. Implement Linux platform for now.

PR #2, opened 2018-10-03
Creating enable_service role

rcabral-splunk:
issue #1 - Creating enable_service task in splunk_common.


roles/splunk_common/tasks/enable_service.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Enable service via boot-start
+  command: "{{ splunk.exec }} enable boot-start -user splunk"
```
nwang92: I would use {{ splunk.user }} here instead of hard-coding this user

PR #6, opened 2018-10-08
Added a validation for a default splunk password

arctan5x:
There is a PR open against the Splunk Docker repo in order to complete this change. We are producing a warning message for an user password that does not conform with the Splunk default password policy.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -303,6 +303,13 @@ def prep_for_yaml_out(inventory):
         del inventory_to_dump["splunk"]["shc"]
     return inventory_to_dump
 
+def validate_splunk_default_password(inventory):
+    password = inventory["all"]["vars"]["splunk"]["password"]
+    if password and len(password) >= 8:
+        return True
+    return False
+    
+
 def main():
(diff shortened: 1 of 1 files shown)


inventory/environ.py
```diff
@@ -303,6 +303,13 @@ def prep_for_yaml_out(inventory):
         del inventory_to_dump["splunk"]["shc"]
     return inventory_to_dump
 
+def validate_splunk_default_password(inventory):
+    password = inventory["all"]["vars"]["splunk"]["password"]
+    if password and len(password) >= 8:
```
outcoldman: but what if you change this configuration from 8 symbols to 12?

outcoldman: I mean `minPasswordLength ` under https://docs.splunk.com/Documentation/Splunk/7.2.0/Admin/Authenticationconf

lephino: I think it'd be safe to flag this as a warning that's sent to stdout and not actually stopping the installation?

arctan5x:
any comments?

PR #9, opened 2018-10-15
Fixing issues with enabling HEC

nwang92:


outcoldman:
Just a note, I believe the `defaults.yaml` enable HEC by default. Any concerts with that? Should it be disabled by default?

nwang92:
@outcoldman I don't think it's very high risk. Changes made here now require the splunk.hec_disabled as well as splunk.hec_token to be properly defined before it actually gets enabled. Even if the default.yml gets generated to enable HEC, it does require the user to explicitly publish the port before it can be consumed? Otherwise, it's only exposed and available only to the docker network it runs in.

outcoldman:
@nwang92 for some reason it is disabled by default. And I assume most customers will expect it to be disabled by default.

I found a lot of Splunk operators to care about security, they even disable managed port for UF. Having HEC port always enabled, even if you don't use it - can make them suspicious. 

Yes, the port will not be available on the host but will be available in container network, pod network, an overlay network, so other containers will be able to connect to it. 

My suggestion will be to keep the default Splunk configuration.

nwang92:
@outcoldman If you take the changes made in this PR and start a Splunk container without a defaults.yml, HEC will not be enabled unless you explicitly pass in a SPLUNK_HEC_TOKEN param. Otherwise, you can create your defaults.yml and modify it to your own liking. But as-is, the default behavior here is that HEC is not enabled.

If open ports/network security is the concern, then are you suggesting we remove all EXPOSE directives in the Dockerfile? Even the [old image](https://github.com/splunk/docker-splunk-legacy/blob/master/enterprise/Dockerfile#L51) made specific ports available to the container network/pod network/other containers, regardless if a user purposely or accidentally enabled something to listen on that port.

outcoldman:
Sounds good. `EXPOSE` actually does not really do anything, only helpful I believe when you want to do something similar to `--publish-all`

PR #15, opened 2018-11-05
Fixed the wrong reference in peering logic

arctan5x:
Wrong reference to the task result.


arctan5x:
screen shot 2018-11-05 at 3 28 33 pm

Fixes the peering.

PR #18, opened 2018-11-13
Enable smartstore functionality

jeff-an:
- Add Ansible plays to write remote/index config to indexes.conf
- Change environ.py to read smartstore config from default.yml and propagate to Ansible
- Supports multiple indexes using a simple remote

### DIFF
--- a/roles/splunk_cluster_master/tasks/configure_indexes.yml
+++ b/roles/splunk_cluster_master/tasks/configure_indexes.yml
@@ -30,6 +30,7 @@
     - { key: "remote.s3.secret_key", value: "{{ index.s3.secret_key }}" }
     - { key: "remote.s3.endpoint", value: "{{ index.s3.endpoint }}" }
   when: index.s3 is defined
+  no_log: true
         
 - name: Write paths for the index
   ini_file:
--- a/roles/splunk_cluster_master/tasks/configure_indexes.yml
+++ b/roles/splunk_cluster_master/tasks/configure_indexes.yml
@@ -2,7 +2,7 @@
 - name: Add stanza for the index specified
   ini_file:
     dest: "{{ splunk.home }}/etc/master-apps/_cluster/local/indexes.conf"
-    section: "{{ index.index }}"
+    section: "{{ index.indexName }}"
     option: "{{ item.key }}"
     value: "{{ item.value }}"
   with_items:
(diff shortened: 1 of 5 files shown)


roles/splunk_cluster_master/tasks/enable_smartstore.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Create indexes.conf conditionally
+  copy:
```
nwang92: I'd go with the `file` module here if all you're doing is making sure a file needs to exist? Otherwise when this re-runs, you're constantly rewriting this indexes.conf

roles/splunk_cluster_master/tasks/configure_indexes.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: Add stanza for the index specified
+  ini_file:
+    dest: "{{ splunk.home }}/etc/master-apps/_cluster/local/indexes.conf"
+    section: "{{ index.index }}"
```
nwang92: NIT but this seems a bit confusing. Maybe `index.name` or `index.indexName` would be more explicit?

roles/splunk_cluster_master/tasks/configure_indexes.yml
```diff
@@ -0,0 +1,45 @@
...
+    - { key: "storageType", value: "remote" }
+    - { key: "path", value: "{{ index.scheme }}://{{ index.remoteLocation }}" }
+
+- name: Write AWS credentials for the remote specified
+  ini_file:
+    dest: "{{ splunk.home }}/etc/master-apps/_cluster/local/indexes.conf"
+    section: "volume:{{ index.remoteName }}"
+    option: "{{ item.key }}"
+    value: "{{ item.value }}"
+  with_items:
```
nwang92: I don't think we've tested this yet, but I think it would be nice to support native IAM access (without plaintext credentials being exposed). 

For this case, I would assume:
1) The cluster is already running in AWS
2) The worker nodes have the correct IAM privileges to access the S3 bucket/volume
3) The containers/pods have access to the IAM privileges (via host networking or other proxying means)

roles/splunk_cluster_master/tasks/enable_smartstore.yml
jeff-an: Actually I can delete the block because ini_file creates the file if it doesn't exist already.

roles/splunk_cluster_master/tasks/configure_indexes.yml
jeff-an: Fixed.

roles/splunk_cluster_master/tasks/configure_indexes.yml
```diff
@@ -0,0 +1,45 @@
...
+    - { key: "remote.s3.access_key", value: "{{ index.s3.access_key }}" }
+    - { key: "remote.s3.secret_key", value: "{{ index.s3.secret_key }}" }
```
nwang92: Another thing I just thought of here: when this runs, do the AWS credentials get printed out to the container logs? If so, we should add loop-control around this section so that we don't print plaintext credentials.

PR #24, opened 2018-11-27
Jackm/timeout

jmeixensperger:
- Use requests in place of urllib2
- Add timeout/delay/retry logic to url install


jmervine:
@lephino +1 thanks!

PR #25, opened 2018-11-27
Fix failure condition for cluster bundle distribution

jeff-an:
Change the "failure" condition to check if the cluster bundle did NOT already exist and the normal stdout terms "applying, bundle" are not present.

Change the "changed_when" condition to check the same.

### DIFF
--- a/roles/splunk_cluster_master/tasks/main.yml
+++ b/roles/splunk_cluster_master/tasks/main.yml
@@ -49,12 +49,11 @@
   command: "{{ splunk.exec }} apply cluster-bundle -auth admin:{{ splunk.password }} --skip-validation --answer-yes"
   register: splunk_cluster_bundle_result
   failed_when: >
-    ("No new bundle will be pushed" not in splunk_cluster_bundle_result.stderr)
-    and ("Rolling restart of the peers" not in splunk_cluster_bundle_result.stderr)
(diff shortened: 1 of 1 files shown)


roles/splunk_cluster_master/tasks/main.yml
```diff
@@ -49,12 +49,11 @@
...
   register: splunk_cluster_bundle_result
   failed_when: >
-    ("No new bundle will be pushed" not in splunk_cluster_bundle_result.stderr)
-    and ("Rolling restart of the peers" not in splunk_cluster_bundle_result.stderr)
-    and ("Applying bundle" not in splunk_cluster_bundle_result.stdout)
-    and ("Applying new" not in splunk_cluster_bundle_result.stdout)
-  changed_when:
-    - splunk_cluster_bundle_result.stdout.find('Applying') != -1
-    - splunk_cluster_bundle_result.stdout.find('bundle') != -1
+    ((splunk_cluster_bundle_result.stdout is not search("[Aa]pply")
```
bb03: This logic is a little difficult to read and reason about, I think a comment here detailing the different conditions that could be encountered (which were clearer from the previous code) would be helpful so that bugs don't get introduced down the line.

jeff-an: Apparently I can't add comments within the task so I added a line above the task.

PR #26, opened 2018-11-27
Task/check first run

dLabsMarijnB:
- Check for first run
- Make install separate task
- Cleanup installer
- Make user seed separate task
- Cleanup user seed after startup
- Only apply license upon first install.

This allows for a more speedy restart of a container.
In addition to this. The user seed file normally gets removed after starting Splunk. However if Splunk gets restarted it will leave any user seed file. Leaving the admin password exposed through the /conf-user-seed REST API.


lephino:
This made me have an idea.  Right now upgrade is currently triggered with a ENV var, by doing something similar and renaming the installer at the end of the process, we could effectively search for first time run, and installer exists, set fact SPLUNK_UPGRADE.  Thoughts?  Like the more manual control, or would you want it automatic?

dLabsMarijnB:
Initially I had made a change that included checking if it was an upgrade.
By checking if the installer exists and if splunkd exists, this would mean it is an upgrade.

But then I figured... why even bother?
The whole concept of upgrade is meaningless.

What is different between installing a new version on top of an existing install and an upgrade?
The only difference I can think of, is doing less, not more.

PR #29, opened 2018-11-29
Feature/cascading config

dLabsMarijnB:
Feature that allows for cascading configuration.
The order it will load the config is:
- splunk_defaults.yml
- /tmp/defaults/default.yml
- Download from the ENV var specified by default to be: SPLUNK_DEFAULTS_URL

Afterwards in the dynamic inventory it will check if /.dockerenv exists, if so, add localhost and try to pull from the url (which should be specified in one of the above configurations) and put this in the hostvars

For example:
```
config:
    host:
        url: https://<server>/{hostname}.yml
        headers:
            TOKEN: <token>
        verify: True
```

Using these parameters you can allow for authenticated, cert verified or not, download of a dynamic link.

For example:
This would allow to specify a deploymentserver in de defaults.
But for the deploymentserver, override it to null, thus all but the deploymentserver will get this setting.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -22,9 +22,12 @@ import random
 import string
 import requests
 import sys
-import urllib2
+import urllib2, urllib3
 import yaml
 from time import sleep
+import ssl
+
+urllib3.disable_warnings()
 
 HERE = os.path.dirname(os.path.normpath(__file__))
 PLATFORM = platform.platform().lower()
@@ -63,6 +66,13 @@ reNamePattern = r"${envPrefix}(.*)"
 inventory = {
     "_meta": {
         "hostvars": {}
+    },
+    "all": {
+        "hosts": [],
+        "children": ["ungrouped"]
+    },
+    "ungrouped": {
+        "hosts": []
     }
 }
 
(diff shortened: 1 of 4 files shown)


dLabsMarijnB:
Seems the identing of the above yaml got messed up.
`config:
    host:
        url: https://<server>/{hostname}.yml
        headers:
            TOKEN: <token>
        verify: True`

inventory/splunk_defaults.yml
```diff
@@ -0,0 +1,68 @@
...
+        url:
+        headers:
+        verify: True
+
+splunk:
+    opt: &opt "/opt"
+    home: &home !!python/object/apply:string.join [[*opt, "/splunk"],'']
+    user: "splunk"
+    group: "splunk"
+    exec: !!python/object/apply:string.join [[*home, "/bin/splunk"], '']
```
lephino: Won't this break on windows?

inventory/environ.py
```diff
@@ -126,150 +132,114 @@ def getSplunkApps(inv):
...
-    vars_scope["splunk"]["idxc"]["search_factor"] = getValueFromDEFAULTS("splunk.idxc.search_factor", casting=int) or 3
-    vars_scope["splunk"]["idxc"]["replication_factor"] = getValueFromDEFAULTS("splunk.idxc.replication_factor", casting=int) or 3
-    vars_scope["splunk"]["idxc"]["replication_port"] = getValueFromDEFAULTS("splunk.idxc.replication_port", casting=int) or 4001
-    vars_scope["splunk"]["enable_service"] = os.environ.get('SPLUNK_ENABLE_SERVICE') or False
-    vars_scope["splunk"]["smartstore"] = getValueFromDEFAULTS("splunk.smartstore")
-
-    vars_scope["splunk"]["app_paths"] = {}
-    if PLATFORM == "linux":
-        vars_scope["splunk"]["group"] = getValueFromDEFAULTS("splunk.group") or "splunk"
-        vars_scope["splunk"]["exec"] = getValueFromDEFAULTS(
```
lephino: The main reasons for these being defined in here is to have python handle the pathing for different OS's.  My only real concern is windows support.

dLabsMarijnB:
Maybe the simplest and cleanest solution would be to put the platform variable inside of the splunk_defaults.yml file.
e.g.
splunk_defaults_linux.yml & splunk_defaults_windows.yml
Keeps it simple and maintance a good overview of what goes where.

Thoughts?

dLabsMarijnB:
There we go, now all the urls can be platform specific as well.

PR #30, opened 2018-11-29
Task/add generic conf

dLabsMarijnB:
This task allows for any generic configuration file to be added, using any generic section and settings.
Combined with the cascaded configuration in PR #29 this virtually allows any scenario to be covered.

For example:
```
splunk:
    conf:
        deploymentclient:
            target-broker:deploymentServer:
                targetUri: "<deploymentServer>:8089"
        inputs:
            monitor:///var/log/messages:
                disabled: False
                index: main
```


dLabsMarijnB:
Absolutely, however editing the conf files is the only option if you do not create a local admin account.
Our standard setup has no local accounts that can be used for authentication.
Only ldap or saml. Meaning as far as I know, there is no way to use cli/rest from ansible.

PR #31, opened 2018-11-30
Adding the ability to run PRE/POST installation tasks

lephino:


site.yml
```diff
@@ -1,7 +1,13 @@
 ---
-- hosts: localhost
+- name: Run Default Installation Plays
+  hosts: localhost
   gather_facts: true
   tasks:
+    - name: Checking for Pre-installation Ansible Tasks
+      include_tasks: "{{lookup('first_found', ansible_pre_tasks)}}"
```
jmervine: I'm a short on Ansible knowledge, so pardon if I'm missed something, but from basic googleing, it doesn't appear that this will work with a URL. If we're going to avoid creating our own image, then this probably will need to support that.

lephino: This looks for the file on disk, if we want to handle url, I can add a download task for it.

lephino: I expected people to mount in the volume their own custom tasks to extend, making it function like the others seems like a good idea.

jmervine: Using `url_get` works either way, as it supports `file:///`, that's what I'd recommend. Again though, I'm an Ansible novice.

jmervine: I realized that in k8s, I can mount these tasks in from a configmap. I'm playing with that now, but it should work without issue.

lephino: added in some download tasks, going to make it take a list and should function well.

lephino: K think this is ready for re-review

inventory/environ.py
```diff
@@ -203,9 +203,10 @@ def loadDefaultSplunkVariables():
         loaded_yaml = yaml.load(yaml_file)
     
     ### Load the defaults for the environment
```
jmervine: Oy, am I missing something or is this:
```
    vars_scope["ansible_pre_tasks"] = os.environ.get("SPLUNK_ANSIBLE_PRE_TASKS") or getValueFromDEFAULTS("ansible_pre_tasks")
    vars_scope["ansible_post_tasks"] = os.environ.get("SPLUNK_ANSIBLE_POST_TASKS") or getValueFromDEFAULTS("ansible_post_tasks")
```
missing from the final commit.

lephino: think I lost it during the merge, good catch.

PR #41, opened 2018-12-11
Fix license install behavior

jmeixensperger:
These changes are supported by additional orca changes (will update with link to PR), so do not merge yet.

The ansible plays were searching the wrong path for a license to apply to the splunk instance. Orca now dumps to the same path for default and local licenses before applying it to splunk.

### DIFF
--- a/roles/splunk_common/tasks/add_splunk_license.yml
+++ b/roles/splunk_common/tasks/add_splunk_license.yml
@@ -1,10 +1,20 @@
 ---
+- name: Check /tmp for license
+  find:
+    paths: "/tmp"
+    patterns: "{{splunk.license_uri}}"
+  register: local_license_file
+  when:
+    - splunk.license_uri is defined
+    - splunk.license_uri is not search("^http.*")
+
 - name: Download Splunk license
   get_url:
     url: "{{ splunk.license_uri }}"
     dest: /tmp/splunk.lic
     timeout: 20
   when:
+    - local_license_file is not defined
     - splunk.license_uri is defined
     - splunk.license_uri is search("^http.*")
     - splunk.role == "splunk_license_master" or not splunk.license_master_included
@@ -13,31 +23,13 @@
   retries: 5
   delay: 3
 
-- name: Set downloaded license location
-  set_fact:
-    splunk_license_local_path: /tmp/splunk.lic
-  when:
-    - splunk.license_uri is defined
-    - splunk.license_uri is search("^http.*")
-    - download_lic is success
-    - splunk.role == "splunk_license_master" or not splunk.license_master_included
-
-- name: Set local license location
-  set_fact:
-    splunk_license_local_path: "{{ splunk.license_uri }}"
-  when:
-    - splunk_license_local_path is not defined
-    - splunk.role == "splunk_license_master" or not splunk.license_master_included
-
 - name: Apply Splunk license
-  command: "{{ splunk.exec }} add licenses {{ splunk_license_local_path }} -auth admin:{{ splunk.password }}"
+  command: "{{ splunk.exec }} add licenses /tmp/splunk.lic -auth admin:{{ splunk.password }}"
   become: yes
   become_user: "{{ splunk.user }}"
   when:
     - splunk.license_uri is defined
-    - splunk_license_local_path is defined and splunk_license_local_path != ""
     - splunk.role == "splunk_license_master" or not splunk.license_master_included
-  ignore_errors: true
   notify:
     - Restart the splunkd service
 
--- a/roles/splunk_common/tasks/add_splunk_license.yml
+++ b/roles/splunk_common/tasks/add_splunk_license.yml
@@ -14,7 +14,7 @@
     dest: /tmp/splunk.lic
     timeout: 20
   when:
-    - local_license_file is not defined
+    - local_license_file.matched == 1
     - splunk.license_uri is defined
     - splunk.license_uri is search("^http.*")
     - splunk.role == "splunk_license_master" or not splunk.license_master_included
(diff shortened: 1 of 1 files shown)


roles/splunk_common/tasks/add_splunk_license.yml
```diff
@@ -13,31 +23,13 @@
   retries: 5
   delay: 3
```
lephino: Probably should do this how I do the "pre-play / post-play" and do a first found lookup, and download to a set location.  Don't forget the nofail on the download incase there's a file path.

jmervine:
Mind rebasing this branch with develop? I'm still getting the following when building off it (but I'm pretty sure that's fixed on develop).

```
TASK [splunk_common : Install Splunk] ******************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'result.rc == 0' failed. The error was: error while evaluating conditional (result.rc == 0): 'dict object' has no attribute 'rc'"}
        to retry, use: --limit @/tmp/ansible-retry/site.retry
```

jmervine:
Thank you for rebasing. I'm getting the following error and the license doesn't appear to be setting up correctly as a result.
```
TASK [splunk_common : Apply Splunk license] ************************************
fatal: [localhost]: FAILED! => {"msg": "The field 'vars' has an invalid value, which includes an undefined variable. The error was: 'dict object' has no attribute 'nfr_license'\n\nThe error appears to
have been in '/opt/ansible/roles/splunk_common/tasks/add_splunk_license.yml': line 15, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears t
o be:\n\n\n- name: Apply Splunk license\n  ^ here\n"}
...ignoring
```

roles/splunk_common/tasks/add_splunk_license.yml
```diff
@@ -5,39 +5,25 @@
...
   when:
-    - splunk.license_uri is defined
-    - splunk_license_local_path is defined and splunk_license_local_path != ""
     - splunk.role == "splunk_license_master" or not splunk.license_master_included
   ignore_errors: true
+  vars:
+    license_locations:
+      - "{{ splunk.license_uri }}"
+      - "{% if downloaded_license_path.dest is defined %}{{downloaded_license_path.dest}}{% endif %}"
+      - "{{ splunk.nfr_license }}"
```
jmervine: This doesn't appear to be defined anywhere in this PR.

jmeixensperger: Did you test this with my changes to the orca repo (linked in original comment)? It included the definition for splunk.nfr_license

jmervine: I missed that. Apologies. I’ll check that out tomorrow.

jmeixensperger: No worries, I checked and that issue still exists, so I didn't define it correctly. I'll append the fix.

inventory/environ.py
```diff
@@ -96,6 +96,8 @@ def getDefaultVars():
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
     defaultVars["splunk"]["license_uri"] = os.environ.get('SPLUNK_LICENSE_URI', '')
+    defaultVars["splunk"]["nfr_license"] = "/tmp/nfr_enterprise.lic"
```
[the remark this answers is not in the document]
jmeixensperger: Changed to retrieve from orca.conf file like the other vars

inventory/environ.py
```diff
@@ -95,7 +95,10 @@ def getDefaultVars():
     defaultVars["splunk"]["deployer_included"] = True if os.environ.get('SPLUNK_DEPLOYER_URL', False) else False
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
-    defaultVars["splunk"]["license_uri"] = os.environ.get('SPLUNK_LICENSE_URI', '')
+    defaultVars["splunk"]["license_uri"] = os.environ.get('SPLUNK_LICENSE_URI', None)
+    defaultVars["splunk"]["nfr_license"] = os.environ.get('SPLUNK_NFR_LICENSE', '/tmp/nfr_enterprise.yml')
```
arctan5x: license is .yml?

jmeixensperger:
I built my own image using docker-splunk and configured my orca branch to use that image instead. I ran each case (ignoring license, local license file, local license download) on a standalone deployment and checked that the expected license was installed correctly.

roles/splunk_common/tasks/add_splunk_license.yml
```diff
@@ -2,42 +2,28 @@
 - name: Download Splunk license
   get_url:
     url: "{{ splunk.license_uri }}"
-    dest: /tmp/splunk.lic
+    dest: "{{ splunk.license_download_dest }}"
     timeout: 20
   when:
-    - splunk.license_uri is defined
-    - splunk.license_uri is search("^http.*")
```
nwang92: You might need this condition added back in? If `splunk.license_uri` is null, the get_url would fail?

jmeixensperger: The get_url is only used when the user specifies a http link with --splunk-local-license. If it fails, the lookup further down will simply skip the download location and revert to the default (nfr) license.

PR #45, opened 2018-12-14
Fixing idempotency of bootstrap cmd

nwang92:


roles/splunk_search_head/tasks/search_head_clustering.yml
```diff
@@ -43,7 +43,9 @@
   when:
     - splunk.role == "splunk_search_head_captain"
   register: task_result
-  until: task_result.rc == 0
+  until: task_result.rc == 0 or task_result.rc == 22
```
jmeixensperger: Be careful using .rc because it won't always be created inside task_result. I believe "until: task_result is success" or "until: task_result|succeeded" will work.

nwang92: Think this should be safe, as long as bin/splunk returns a proper exit code. task_result is success won't necessarily work here as one of the break conditions is a failure event (which isn't necessarily a failure).

PR #60, opened 2019-01-04
Java support

arctan5x:
Added a capability to install openjdk8 or oracle8


roles/splunk_common/tasks/java_tasks/install_oracle11_jdk.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Download the Java distribution
+  command: "wget --header 'Cookie: oraclelicense=accept-securebackup-cookie' https://download.oracle.com/otn-pub/java/jdk/11.0.1+13/90cf5d8f270a4347a95050320eef3fb7/jdk-11.0.1_linux-x64_bin.tar.gz -P /opt/container_artifact"
```
bb03: The download link would be good in a var, just in case the path changes

arctan5x: we can switch out the link as a variable, but the user also has to pass in a tar file name and uncompressed file name in order to run next tasks in this playbook. Would it be a considerable value to the customer?

PR #61, opened 2019-01-07
Use the correct indexer search/replication factor

HetengZhang:
Use the correct indexer search/replication factor in splunk_cluster_master/tasks/main.yml

### DIFF
--- a/roles/splunk_cluster_master/tasks/main.yml
+++ b/roles/splunk_cluster_master/tasks/main.yml
@@ -1,16 +1,4 @@
 ---
-- name: Get indexer count
-  set_fact:
-    num_indexer_hosts: "{{ groups['splunk_indexer'] | length }}"
-
-- name: Lower indexer search/replication factor
-  set_fact:
-    splunk:
-      idxc:
-        search_factor: 1
-        replication_factor: 1
-  when: num_indexer_hosts|int < 3
-
 - name: Set indexer discovery
   uri:
     url: "https://127.0.0.1:{{ splunk.svc_port }}/servicesNS/nobody/system/configs/conf-server"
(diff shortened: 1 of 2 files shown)


roles/splunk_cluster_master/tasks/main.yml
```diff
@@ -35,7 +35,7 @@
   when: splunk.smartstore
 
 - name: Set the current node as a Splunk indexer cluster master
-  command: "{{ splunk.exec }} edit cluster-config -mode master -replication_factor {{ splunk.idxc.replication_factor }} -search_factor {{ splunk.idxc.search_factor }} -secret '{{ splunk.idxc.secret }}' -cluster_label '{{ splunk.idxc.label }}' -auth 'admin:{{ splunk.password }}'"
```
arctan5x: In fact, could we not define idxc_replication_factor and idxc_search_factor up in this playbook since they are just redefined values of splunk.idx.{{variable}}.

HetengZhang: When num_indexer_hosts < 3, they will be set to 1.

lephino:
This should be changed in the user's defaults.yml file, and not left up to ansible to auto-figure out.  There's no need for this change, and can still be accomplished by using either the env var or default var

lephino:
Rethinking on this, happy to have this just run as a 2nd task, just set the first one back.  There's nothing wrong with setting it to defaults.yml and then dropping for the user if the count dropped below 3.

arctan5x:
I think it is always simpler to enforce a change in one place. default.yml already has these variables to be passed in. If you necessarily need to override it, please have this variable overriding done in the environ.py. Ansible should merely just work with the variables provided by the inventory script.

HetengZhang:
Lower idxc factors in environ.py now.

PR #63, opened 2019-01-09
Adding handler to restart after app installation

nwang92:
Addressing https://github.com/splunk/docker-splunk/issues/92


nwang92:
@lephino I added a separate task to check for any required restarts. As long as the app has a `state_change_requires_restart` or special triggers, it should notify the `/services/messages/restart_required` which will prompt the handler to run at the end of the playbook execution.

PR #64, opened 2019-01-09
Dynamic mode + Forwarder builds

jmeixensperger:
- Added environment vars to handle enterprise/uf builds at the same time
- Fixed bug where search heads failed to register peer indexers

- This PR depends on related changes to the orca repo: https://git.splunk.com/projects/TOOLS/repos/orca/pull-requests/1559/overview

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -95,6 +95,7 @@ def getDefaultVars():
     defaultVars["splunk"]["deployer_included"] = True if os.environ.get('SPLUNK_DEPLOYER_URL', False) else False
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
+    defaultVars["splunk"]["dynamic"] = os.environ.get('SPLUNK_DYNAMIC', False)
     defaultVars["splunk"]["license_uri"] = os.environ.get('SPLUNK_LICENSE_URI', None)
     defaultVars["splunk"]["nfr_license"] = os.environ.get('SPLUNK_NFR_LICENSE', '/tmp/nfr_enterprise.lic')
     defaultVars["splunk"]["ignore_license"] = os.environ.get('SPLUNK_IGNORE_LICENSE', False)
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -95,7 +95,6 @@ def getDefaultVars():
     defaultVars["splunk"]["deployer_included"] = True if os.environ.get('SPLUNK_DEPLOYER_URL', False) else False
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
-    defaultVars["splunk"]["dynamic"] = os.environ.get('SPLUNK_DYNAMIC', False)
     defaultVars["splunk"]["license_uri"] = os.environ.get('SPLUNK_LICENSE_URI', None)
     defaultVars["splunk"]["nfr_license"] = os.environ.get('SPLUNK_NFR_LICENSE', '/tmp/nfr_enterprise.lic')
     defaultVars["splunk"]["ignore_license"] = os.environ.get('SPLUNK_IGNORE_LICENSE', False)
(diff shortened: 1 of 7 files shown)


roles/splunk_common/tasks/install_splunk.yml
```diff
@@ -1,5 +1,20 @@
 ---
-- name: Install Splunk
+- name: Download splunk package
```
arctan5x: remote_src variable in the task below controls if the src is remote and needs to be fetched so I think this task might be doing a repetitive job.

roles/splunk_common/tasks/install_splunk.yml
```diff
@@ -22,7 +37,7 @@
 
 - name: Remove installer
   file:
-    dest: "{{ splunk.build_location }}"
+    dest: "/tmp/splunk.tgz"
```
arctan5x: This destination still is required to be a flexible variable since Orca sends Splunk distribution into /opt/container_artifact directory.

roles/splunk_common/tasks/install_splunk.yml
```diff
@@ -9,11 +24,11 @@
   when: ansible_system is match("Linux")
   register: result
   until: result is succeeded
-  retries: "{{ retry_num }}"
+  retries: {{ retry_num }}
   delay: 3
 
 - name: Install Splunk (Windows)
-  command: "msiexec /I {{ splunk.build_location }} AGREETOLICENSE=yes LAUNCHSPLUNK=0 INSTALLDIR=C:\\\\opt\\\\splunk /passive /qn"
+  command: "msiexec /I /tmp/splunk.tgz AGREETOLICENSE=yes LAUNCHSPLUNK=0 INSTALLDIR=C:\\\\opt\\\\splunk /passive /qn"
```
arctan5x: same here regarding the path /tmp/splunk.tgz

inventory/environ.py
```diff
@@ -95,6 +95,7 @@ def getDefaultVars():
     defaultVars["splunk"]["deployer_included"] = True if os.environ.get('SPLUNK_DEPLOYER_URL', False) else False
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
+    defaultVars["splunk"]["dynamic"] = os.environ.get('SPLUNK_DYNAMIC', False)
```
arctan5x: Traditional logic is that if the build location starts with http, we treat the source as remote. How is Dynamic changing this behavior?

roles/splunk_common/tasks/install_splunk.yml
jmeixensperger: Yes, I made this task for windows (check the "when:") because it doesn't use remote_src.

inventory/splunk_defaults_linux.yml
```diff
@@ -21,6 +21,7 @@ config:
 
 splunk:
     build_location: "/tmp/splunk.tgz"
+    uf_build_location: "/tmp/splunk.tgz"
```
arctan5x: i think these are ok to be default values. Though these variables should be overwritable in environ.py

roles/splunk_universal_forwarder/tasks/main.yml
```diff
@@ -1,15 +1,34 @@
...
+    remote_src: "{{ splunk.uf_build_remote_src }}"
+  register: install_result
+  until: install_result is succeeded
+  retries: 5
+  delay: 3
   when: ansible_system is match("Linux")
 
 - name: Install Splunk universal forwarder (Windows)
-  command: "msiexec /I {{ splunk.build_location }} AGREETOLICENSE=yes LAUNCHSPLUNK=0 INSTALLDIR=C:\\\\opt\\\\splunkforwarder /passive /qn"
+  command: "msiexec /I /tmp/splunk.tgz AGREETOLICENSE=yes LAUNCHSPLUNK=0 INSTALLDIR=C:\\\\opt\\\\splunkforwarder /passive /qn"
```
arctan5x: same thing about hardcoded splunk distribution path.

roles/splunk_common/tasks/install_splunk.yml
jmeixensperger: The problem with using "splunk.build_location" is that it can become a local build path or a custom build url. Otherwise, it will always be "/tmp/splunk.tgz"

inventory/environ.py
jmeixensperger: It doesn't... I forget what I was using it for, but we don't need it.

inventory/splunk_defaults_linux.yml
jmeixensperger: Yes, they are overwritten by local builds and custom build url's.

roles/splunk_common/tasks/install_splunk.yml
```diff
@@ -1,30 +1,48 @@
 ---
-- name: Install Splunk
+- name: Download Slunk package (Windows)
```
JasonXuSplunk: Typo: Splunk

PR #66, opened 2019-01-15
Fix key error

HetengZhang:


inventory/environ.py
```diff
@@ -104,7 +104,7 @@ def getDefaultVars():
     defaultVars["hide_password"] = True if os.environ.get('HIDE_PASSWORD', "").lower() == "true" else False
 
     # Lower indexer search/replication factor when indexer hosts less than 3
-    if len(inventory["splunk_indexer"]["hosts"]) < 3:
+    if inventory.has_key("splunk_indexer") and len(inventory["splunk_indexer"]["hosts"]) < 3:
```
arctan5x: would you do a one more nested check of the key "hosts"?

PR #68, opened 2019-01-15
Bugfix/https apps

nwang92:
This PR should address the following:
* Handle app installation from HTTPS urls (previously only supported splunkbase + HTTP links)
* Combining functionality of `SPLUNK_APPS_URL` and `SPLUNK_LOCAL_APPS`
* Installation of apps in distributed environments (deployers push to search heads, cluster master pushes to indexers)
* Support installation of premium apps, including ES and ITSI
* Fixing bug with setting pass4SymmKey + shcluster_label via REST (I don't think using REST works in this case)
* Fixing bug with log forwarding on license master + cluster master

Some to-do's:
* Filter out which apps get pushed via shcluster-bundle and cluster-bundle


nwang92:
Tentatively going to merge this in, I'm fixing one last bit to optimize cluster-bundle pushes from the CM.

PR #69, opened 2019-01-16
Bugfix/backwards compatibility

nwang92:
Patching how licensing is done to preserve backwards compatibility.

Previously, we only had `SPLUNK_LICENSE_URI` - now we have `SPLUNK_LICENSE_URI`, `SPLUNK_NFR_LICENSE`, and `SPLUNK_LICENSE_INSTALL_PATH`. This ended up breaking some images that based off splunk/splunk, as `SPLUNK_LICENSE_URI` doesn't handle normal file paths (/path/to/splunk.lic) anymore.


roles/splunk_common/tasks/add_splunk_license.yml
```diff
@@ -6,6 +6,7 @@
     timeout: 20
   when:
     - splunk.role == "splunk_license_master" or not splunk.license_master_included
+    - splunk.license_uri is match("^(http[s]|file)://.*")
```
mikedickey: Just noticing.. shouldn't this be something like `match("^(http|https|file)://.*")` or `match("^(https?|file)://.*")`?

nwang92: Yeah that would be a better rex, I'll amend this

PR #70, opened 2019-01-16
Removing the undocumented splunk_search_head_captain role

mikedickey:
Intead, we now check to see if SPLUNK_SEARCH_HEAD_CAPTAIN_URL matches the current hostname.
If so, it bootstraps as the captain.


nwang92:
This change might break a lot of the current scenarios (compose + manifests) in docker-splunk, at least anything that uses `SPLUNK_ROLE=splunk_search_head_captain`. 

Do we want to phase this out gradually? We can incorporate this change + support backwards compatibility by creating a `roles/splunk_search_head_captain` with no tasks, but calls a dependency on `role: splunk_search_head`. This should give users one release cycle to switch over to the new format, after which we can remove everything related to `splunk_search_head_captain`.

nwang92:
Opened https://github.com/splunk/splunk-ansible/pull/71 for my proposal

PR #71, opened 2019-01-17
Adding dummy role to support backwards compatibility

nwang92:


mikedickey:
Only comment is we may want to modify the when checks for the splunk_search_head_captain fact in site.yml, so that it is true either the URL==host OR role==splunk_search_head_captain

nwang92:
@mikedickey done - although thinking about it some more, it might not be necessary? If you define a particular container as a the captain but CAPTAIN_URL!=host, it's likely the deployment won't be brought up successfully anyways. But the additional security blanket here likely won't hurt either so 🤷‍♂️

site.yml
```diff
@@ -12,7 +12,7 @@
     - set_fact:
         splunk_search_head_captain: true
       when:
-        - actual_hostname.stdout == lookup('env', 'SPLUNK_SEARCH_HEAD_CAPTAIN_URL')
+        - actual_hostname.stdout == lookup('env', 'SPLUNK_SEARCH_HEAD_CAPTAIN_URL') or splunk.role == "splunk_search_head_captain"
```
mikedickey: I think this also requires a similar (but inverse) "and ... !=" in the task below, so it doesn't get flipped to false

PR #72, opened 2019-01-17
Suppress errors when ansible_post_tasks is a file path

mikedickey:
Only run tasks "Download pre-setup playbooks" when ansible_post_tasks is a web URL (suppress errors when it is a file path)

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -9,6 +9,7 @@
       when:
         - ansible_pre_tasks is defined
         - ansible_pre_tasks is not none
+        - ansible_pre_tasks is match("^(http|https|file)://.*")
         - ansible_system is match("Linux")
       get_url:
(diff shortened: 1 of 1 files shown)


site.yml
```diff
@@ -72,6 +72,7 @@
         - ansible_post_tasks is defined
         - ansible_post_tasks is not none
         - ansible_system is match("Linux")
+        - ansible_post_tasks is match("^(http|https)://.*")
```
nwang92: Should we include `file://` too?

mikedickey: I think if it's a local file, we also do not want it to run download tasks.

lephino: I think we will if it's file:// it'll move it to the right name

PR #75, opened 2019-01-24
Update oracle 8 link

gracening:
Updated the link for downloading Java Oracle 8. The old one stopped working and gets 404 error. You can see this when doing `make sample-compose-up` with the `1so_java.yaml` test scenario in docker-splunk.

### DIFF
--- a/roles/splunk_common/tasks/java_tasks/install_oracle8_jdk.yml
+++ b/roles/splunk_common/tasks/java_tasks/install_oracle8_jdk.yml
@@ -1,9 +1,9 @@
 ---
 - name: Download the Java distribution
   get_url:
-    url: https://download.oracle.com/otn-pub/java/jdk/8u191-b12/2787e4a523244c269598db4e85c51e0c/jdk-8u191-linux-x64.tar.gz
+    url: https://download.oracle.com/otn-pub/java/jdk/8u201-b09/42970487e3af4f5aa5bca3f542482c60/jdk-8u201-linux-x64.tar.gz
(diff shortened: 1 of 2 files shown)


gracening:
aw heck how do I unassign I misclicked 😢

bb03:
This is something that I expect will change in the future, so I would encourage you to move it into the vars, instead of placing it in the task - this way its easier for a user to replace the value when oracle updates java (instead of needing us to re-release a package)

roles/splunk_common/tasks/java_tasks/install_oracle8_jdk.yml
```diff
@@ -1,9 +1,9 @@
 ---
 - name: Download the Java distribution
   get_url:
-    url: https://download.oracle.com/otn-pub/java/jdk/8u191-b12/2787e4a523244c269598db4e85c51e0c/jdk-8u191-linux-x64.tar.gz
+    url: "{{ java_oracle_url }}"
```
nwang92: Wonder if we should add something like:
```
retries: 5
delay: 10
``` 
to this section? We may be more failure-tolerant to ephemeral networking problems this way.

nwang92: Maybe also add a `timeout` into the get_url parameter: https://docs.ansible.com/ansible/latest/modules/get_url_module.html

The default is 10s, which might not be conservative enough. Maybe something like 30s would be safe.

PR #81, opened 2019-01-28
Bugfix/consistent forwarding patterns

nwang92:
This should fix https://github.com/splunk/docker-splunk/issues/89

What I did:
* add_forward_server.yml is called within enable_forwarding.yml
* enable_forwarding will handle the case of forwarding to indexer cluster, indexers individually, standalones individually in that order
* UF + HF will now support indexer discovery when used with an indexer cluster master

### DIFF
--- a/roles/splunk_common/tasks/enable_forwarding.yml
+++ b/roles/splunk_common/tasks/enable_forwarding.yml
@@ -1,6 +1,7 @@
 ---
-# Configure forwarding for index-clustered deployments
-- name: Setup indexer_discovery:splunk-indexer stanza in outputs.conf
+# Configure forwarding to indexer cluster master
+# See: https://docs.splunk.com/Documentation/Splunk/latest/Indexer/indexerdiscovery
+- name: Setup indexer discovery for index-clustering
   uri:
     url: "https://127.0.0.1:{{ splunk.svc_port }}/servicesNS/nobody/system/configs/conf-outputs"
     method: POST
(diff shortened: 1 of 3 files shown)


roles/splunk_common/tasks/enable_forwarding.yml
```diff
@@ -16,15 +17,17 @@
...
-- name: Setup tcpout:group1 stanza in outputs.conf for clustered deployments
+- name: Setup tcpout group for index-clustering
   uri:
     url: "https://127.0.0.1:{{ splunk.svc_port }}/servicesNS/nobody/system/configs/conf-outputs"
     method: POST
     user: admin
     password: "{{ splunk.password }}"
     validate_certs: false
-    body: "name=tcpout:group1&indexerDiscovery=splunk-indexer&forwardedindex.0.whitelist=_internal&forwardedindex.1.whitelist=_audit&forwardedindex.filter.disable=false"
+    body: "name=tcpout:group1&indexerDiscovery=splunk-indexer"
```
nwang92: FYI - I took the liberty of removing some of the parameters here. I checked in ${SPLUNK_HOME}/etc/system/local and found the defaults for `[tcpout]` can probably be used in lieu of these rules:
```
[tcpout]
maxQueueSize = auto
forwardedindex.0.whitelist = .*
forwardedindex.1.blacklist = _.*
forwardedindex.2.whitelist = (_audit|_internal|_introspection|_telemetry)
forwardedindex.filter.disable = false
```

PR #83, opened 2019-01-29
Apply shcluster bundle optional

jmervine:
This is too restrictive in it not ignoring errors and fails too quickly, as well.

1. Ignoring errors, as this failing should cause an entire cluster to fail to boot. This can cause race conditions with the deployer, captain and searchers.
2. Using `retry_num` for `retries` and setting `delay: 3` to match other plays within this repo.

### DIFF
--- a/roles/splunk_deployer/tasks/push_apps_to_search_heads.yml
+++ b/roles/splunk_deployer/tasks/push_apps_to_search_heads.yml
@@ -20,3 +20,8 @@
   retries: "{{ retry_num }}"
   delay: 3
   ignore_errors: true
+
+- debug: 
+    msg: "WARNING: Applying shcluster bundle failed - proceeding anyways..."
+  when: 
+    - shcluster_bundle.rc != 0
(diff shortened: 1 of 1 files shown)


lephino:
Kinda agree with @nwang92 but willing to put it to a vote on the community.  Technically a large bundle push on the start could fail (like installing 10 apps) and isn't indicative of a real issue on the cluster.  Those apps should be pre-seeded, which I have as a backlog item to make possible to install.  The question is, as a community, do you want this to stop the ansible plays and call the stack a failure?  Or have them continue and you'll triage the issue once the stack is up?

arctan5x:
I vote for call it a failure and stop ansible plays. My reason is that ansible should always fail when it cannot produce the Splunk environment that does not exactly match to a user's explicit command. We do not want a situation where the ansible plays successfully run but the environment is not complete. This scenario can be misleading.

lephino:
Maybe we add a flag for "ignore_bundle_push_failures", true false?

nwang92:
I'd be ok with something like: https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml where, based on some potential error conditions, we can at least print out a debug message to some end users?

nwang92:
I approved this because I do see the retry limit causing an issue with a new test I added, but happy to continue the discussion re: treating app installations as failures/not failures.

jmervine:
I’d be happy to add a debug message to this PR.

jmervine:
@nwang92 @jmeixensperger Adding debug messaging as discussed.

PR #90, opened 2019-01-31
Added openjdk11 and other improvements

arctan5x:


roles/splunk_common/tasks/java_tasks/install_oracle8_jdk.yml
```diff
@@ -57,3 +57,34 @@
...
+  when:
+    - ansible_distribution == 'Debian'
+
+- name: Set JAVA_HOME in splunk-launch.conf
+  become: yes
+  become_method: sudo
+  become_user: root
+  lineinfile:
+    path: "{{splunk.home}}/etc/splunk-launch.conf"
+    regexp: '^JAVA_HOME'
```
nwang92: If there's no regex match, will it still write this line?

arctan5x: yes

inventory/environ.py
```diff
@@ -104,13 +104,24 @@ def getDefaultVars():
...
-            defaultVars["java_update_version"] = re.search("jdk-8u(\d+)-linux-x64.tar.gz", defaultVars["java_oracle_url"]).group(1)
-        except:
-            raise Exception("Invalid Java download URL format")
+        if java_version == "oracle:8":
+            defaultVars["java_download_url"] = os.environ.get("JAVA_DOWNLOAD_URL", "https://download.oracle.com/otn-pub/java/jdk/8u201-b09/42970487e3af4f5aa5bca3f542482c60/jdk-8u201-linux-x64.tar.gz")
+            try:
+                defaultVars["java_update_version"] = re.search("jdk-8u(\d+)-linux-x64.tar.gz", defaultVars["java_download_url"]).group(1)
+            except:
+                raise Exception("Invalid Java download URL format")
+        elif java_version == "openjdk:11":
```
jmeixensperger: What about openjdk:8? Do we not need to set java_download_url or java_update_version?

arctan5x: For openjdk, we install it through the package manager so no need for the url.

PR #95, opened 2019-02-01
Removed become_user=splunk as the become_user

rcabral-splunk:
Also removed become_user from enable_service task as it requires privileges to update /etc/init.d.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -22,7 +22,6 @@ action_warnings = False
 [privilege_escalation]
 become = Yes
 become_method = sudo
-become_user = splunk
 become_exe = "sudo -u"
 
 
--- a/roles/splunk_common/tasks/enable_service.yml
+++ b/roles/splunk_common/tasks/enable_service.yml
@@ -2,4 +2,3 @@
 - name: Enable service via boot-start
   command: "{{ splunk.exec }} enable boot-start -user {{ splunk.user }}"
   become: yes
-  become_user: "{{ splunk.user }}"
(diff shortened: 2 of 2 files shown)


dLabsMarijnB:
Some additional flags should be added to the enable-service in order to make this work.
When not running any before_start_cmd you need to add:
--accept-license --answer-yes --no-prompt

rcabral-splunk:
> Some additional flags should be added to the enable-service in order to make this work.
> When not running any before_start_cmd you need to add:
> --accept-license --answer-yes --no-prompt

Not quite sure I understood your suggestion. Are you proposing to include those flags to enable-boot? If so, those aren't valid parameters for the boot-start.

lephino:
I think as long as this is merged with 
https://github.com/splunk/splunk-ansible/pull/97, you guys hit the needs.

dLabsMarijnB:
I have just tested it again to verify and make sure.

The order in which things get executed in the playbooks means the enable boot-start happens before anything else. (Unless you use the pre-start execution option, which is not a guarantee)
Then:
- You do need the --accept-license at enable boot-start, otherwise it will hang on the license prompt
- You also need the --no-prompt because it will hang at the password input prompt
- And the --answer-yes is just for good measure.

Most of the changes requested I have incorporated into #97.

rcabral-splunk:
Figured out why it's working. `enable_s2s_port` task runs before `enable_service` and it is already passing those parameters. We don't necessarily need to add those to `enable_service` again, because  if it fails, the whole playbook would also do it so. However, it won'r hurt if we also add those extra args as well. Updating the PR with it.

nwang92:
Seems like https://github.com/splunk/splunk-ansible/pull/97 solves this problem + also adds systemd. Maybe we should close this PR in favor of that?

rcabral-splunk:
> Seems like #97 solves this problem + also adds systemd. Maybe we should close this PR in favor of that?

Correct! Sounds good to me.

lephino:
Closing for 97

PR #97, opened 2019-02-05
Systemd support

dLabsMarijnB:
Added support for starting / stopping / restarting the splunkd based on the method used, being systemd, windows or cli if the enable_service is false.


lephino:
Same here, please fix the conflict or re merge master and remove the tracking

PR #98, opened 2019-02-05
Remove first login

dLabsMarijnB:
By creating the .ui_login file we remove the first login screen.


lephino:
Can you fix the conflict please and we'll get this merged in

PR #103, opened 2019-02-08
Environment for build location

SkyKnightSKS:
Related to https://github.com/splunk/docker-splunk/pull/123
This allows us to maintain the original filename and use the get_fact task to allow ansible to know which version we want to install.


jmervine:
This (in conjunction with https://github.com/splunk/docker-splunk/pull/123) appears to be a fairly large functionality change. Is there a plan to include documentation to support it?

SkyKnightSKS:
How is this such a large functionality change?
The only difference here is to maintain the original filename in order to be able to parse the version of the installer we have available.

jmervine:
I meant the switch to systemd.

SkyKnightSKS:
That would only be in case splunk.enable_service is true.
When that flag is set it will also call the enable boot-start.

As of Splunk 7.2.2 you must use systemd if start/stop/restart after you have enabled boot-start.
To build support for the older init.d scripts we need to be able to recognise which version we are, so that we can use the correct mechanism.

roles/splunk_common/tasks/start_splunk.yml
```diff
@@ -1,20 +1,24 @@
...
-- name: Start Splunk via Windows service
+- name: "Start Splunk via Windows service"
   win_service:
     name: splunkd
     state: restarted
   when: splunk.enable_service and not ansible_system is match("Linux")
+
+- name: "Wait for port 8000 to become open"
+  wait_for:
+    port: 8000
```
nwang92: 8000 isn't open on Splunk Universal Forwarder because there's no web UI. If we want to make this portable, we should switch to 8089.

roles/splunk_common/handlers/main.yml
```diff
@@ -20,9 +20,14 @@
...
   listen: "Restart the splunkd service"
   win_service:
     name: splunkd
     state: restarted
   when: splunk.enable_service and not ansible_system is match("Linux")
+
+- name: "Wait for port 8000 to become open"
+  listen: "Restart the splunkd service"
+  wait_for:
+    port: 8000
```
nwang92: Same here - for consistency across Splunk + UF, we can't use 8000

inventory/environ.py
```diff
@@ -183,6 +183,7 @@ def overrideEnvironmentVars(vars_scope):
         vars_scope["splunk"]["idxc"] = {}
     vars_scope["splunk"]["idxc"]["secret"] = os.environ.get('SPLUNK_IDXC_SECRET', vars_scope["splunk"]["idxc"]["secret"])
     vars_scope["splunk"]["enable_service"] = os.environ.get('SPLUNK_ENABLE_SERVICE', vars_scope["splunk"]["enable_service"])
+    vars_scope["splunk"]["build_location"] = os.environ.get('SPLUNK_INSTALLER', vars_scope["splunk"]["build_location"])
```
nwang92: This works and I think it's safe to leave it. But it should be noted in docs somewhere that `SPLUNK_INSTALLER` and `SPLUNK_BUILD_URL` will write to the same location `vars_scope["splunk"]["build_location"]`, with `SPLUNK_BUILD_URL` taking precedence.

Maybe we can also capture this behavior with the tests/small/test_environ.py

nwang92:
@SkyKnightSKS can you merge in develop and resolve the merge conflict?

Issue #101, opened 2019-02-07
First run check question.


jmervine:
Look at the following...

https://github.com/splunk/splunk-ansible/blob/8149b68198197bbe7c3aa8be1ea30a7f63b1d80e/roles/splunk_common/tasks/main.yml#L9-L12

I'd assume this would prevent upgrades from being applied. E.g. deploying splunk/splunk:7.2.1 and then splunk/splunk:7.2.3 wouldn't actually install the new version of splunk. Is that correct? Is that expected?

lephino:
Think this is due to: https://github.com/splunk/splunk-ansible/pull/100 not being in yet, it specifically calls out upgrade.

lephino:
Actually, think you might be right, I don't see where we include it on upgrade.

lephino:
https://github.com/splunk/splunk-ansible/blob/develop/site.yml#L62

jmervine:
How does `splunk.upgrade` get set here: https://github.com/splunk/splunk-ansible/blob/develop/site.yml#L62-L67?

The only thing one should have to do to upgrade splunk is to deploy a new image.

jmervine:
Gotcha, then this issue might be invalid. If you agree close it.

dLabsMarijnB:
Currently it still considers it a first run based on the fact if the /opt/splunk/bin/splunk is available or not. This should be changed to verify if there is a version installed.
Next would be to change the when clause for the install.

My idea would be to trigger the install_splunk.yml task if:
It is a first run
or
it is an upgrade defined by splunk_upgrade (check if version is different) AND splunk.allow_upgrade (which is set by the user to allow or disallow upgrades)

lephino:
@mikedickey I think this logic is met, and we can close this pr, however your pr directly hits this logic as well.  I'll leave this one to you to close.

mikedickey:
I agree this is solved by changes in SkyKnightSKS's PR, and then tweaked a little bit in my own PR (I already merged those changes in and resolved conflicts).

PR #104, opened 2019-02-08
Adapt upgrade logic

SkyKnightSKS:
https://github.com/splunk/splunk-ansible/issues/101

Changed the upgrade logic to be part of the normal run.
Added more facts allowing to identify the difference between a fresh install, a re-install in docker, the first time we run splunk and when it is an upgrade.
Allowed for a flag to allow or disallow upgrades.
Set the allow_upgrade flag to false by default.


arctan5x:
What are the use cases for upgrade and allow_upgrade? Seems that using both variables can be confusing.

SkyKnightSKS:
The old splunk.upgrade was renamed to splunk.allow_upgrade, which is now used to indicate that an upgrade is allowed or disallowed.

There is an internal fact which is acquired using the get_facts.yml task, named splunk_upgrade. This fact will show if we have a newer installer than the currently installed version.

Both need to be true before an upgrade should occur.

SkyKnightSKS:
splunk.allow_upgrade can be manipulated by environment variables or defaults.

while the splunk_upgrade is detected based on the version and cannot be changed.

inventory/environ.py
```diff
@@ -182,19 +181,13 @@ def overrideEnvironmentVars(vars_scope):
         vars_scope["splunk"]["idxc"] = {}
     vars_scope["splunk"]["idxc"]["secret"] = os.environ.get('SPLUNK_IDXC_SECRET', vars_scope["splunk"]["idxc"]["secret"])
     vars_scope["splunk"]["enable_service"] = os.environ.get('SPLUNK_ENABLE_SERVICE', vars_scope["splunk"]["enable_service"])
+    vars_scope["splunk"]["allow_upgrade"] = os.environ.get('SPLUNK_ALLOW_UPGRADE', vars_scope["splunk"]["allow_upgrade"])
```
jmervine: Pedantic, perhaps, but I think this should default to `True`, perhaps being renamed to `SPLUNK_DISALLOW_UPGRADE`.

SkyKnightSKS: I have personal preference for a positive flag instead of a negative one. But this set aside I agree with you that in an automation system, the upgrade should by default be automatic.
The flag should be kept in order to have the ability to block certain nodes from upgrading while others do.

if to call it allow or disallow... I leave that up to the majority to decide.

mikedickey:
Please see another PR I'm working on here https://github.com/splunk/splunk-ansible/pull/116/files

It's slightly different things, but along similar lines, and may have conflicts with this PR.

PR #105, opened 2019-02-11
Add ssl variables by environment within environ.py

sscholl:
The lack of this functionallity makes it really hard to inject SSL config. I use these files. May you want to add them to the README.

Thanks


nwang92:
How was this tested? If this works for your scenario, I'd like to take your use case and add a test for it on the docker-splunk side.

sscholl:
@nwang92 I used docker-splunk to test it. I set the 4 environment variables and mounted the .pem and .key file by Docker volumes. It worked just right away.

PR #107, opened 2019-02-11
Fixing things that break UF

nwang92:


nwang92:
Root cause was the "Wait for 8000" port, which doesn't exist for UF's at all. I cleaned up some other things along the way.

Issue #106, opened 2019-02-11
Start/stop splunk tasks fail when Linux distro isn't systemd based


rcabral-splunk:
For those systemv based Linux, start/stop tasks fail, since it currently only supports systemd based ones. We also need to support sytemv, for the AMI builds that make use of Amazon Linux (version 1).

nwang92:
I believe @rcabral-splunk had some open issues with this still. Please re-open the PR when you're ready.

SkyKnightSKS:
When using systemctl there is still an issue.  Systemctl fires off the splunkd process and continues.  But Splunk is not yet up and running.  API calls following this will not be able to connect as the process is not available to answer.

There was a wait-for-port task which was removed for UF support, a proper way to handle the wait for process part should be implemented.

rcabral-splunk:
> When using systemctl there is still an issue. Systemctl fires off the splunkd process and continues. But Splunk is not yet up and running. API calls following this will not be able to connect as the process is not available to answer.
> 
> There was a wait-for-port task which was removed for UF support, a proper way to handle the wait for process part should be implemented.

I guess we can close this PR in favor of #115. If you are still having issues with process not able to connect, let's discuss on #115 perhaps?

lephino:
Closing, move discussions to #115

PR #108, opened 2019-02-11
Removing dependency from systemd

nwang92:
Addressing https://github.com/splunk/splunk-ansible/issues/106

### DIFF
--- a/roles/splunk_common/handlers/main.yml
+++ b/roles/splunk_common/handlers/main.yml
@@ -17,6 +17,8 @@
     name: splunk
     state: restarted
   when: splunk.enable_service and ansible_system is match("Linux")
+  become: yes
+  become_user: root
 
 - name: Restart the splunkd service via Windows service
   listen: "Restart the splunkd service"
--- a/roles/splunk_common/handlers/main.yml
+++ b/roles/splunk_common/handlers/main.yml
@@ -13,9 +13,9 @@
 
 - name: Restart the splunkd service via systemd
   listen: "Restart the splunkd service"
-  systemd:
+  service:
+    name: splunk
     state: restarted
-    name: Splunkd
   when: splunk.enable_service and ansible_system is match("Linux")
 
 - name: Restart the splunkd service via Windows service
--- a/roles/splunk_common/tasks/enable_service.yml
+++ b/roles/splunk_common/tasks/enable_service.yml
@@ -2,7 +2,7 @@
 - name: "Enable service via boot-start - Linux"
   become: yes
   become_user: "root"
-  command: "{{ splunk.exec }} enable boot-start -user {{ splunk.user }} --accept-license --answer-yes --no-prompt"
+  command: "{{ splunk.exec }} enable boot-start -user {{ splunk.user }} -systemd-unit-file-name splunk --accept-license --answer-yes --no-prompt"
   when: ansible_system is match("Linux")
 
 - name: "Enable service via boot-start - Windows"
--- a/roles/splunk_common/tasks/enable_service.yml
+++ b/roles/splunk_common/tasks/enable_service.yml
@@ -2,11 +2,9 @@
 - name: "Enable service via boot-start - Linux"
   become: yes
   become_user: "root"
-  command: "{{ splunk.exec }} enable boot-start -user {{ splunk.user }}  --accept-license --answer-yes --no-prompt"
-  args:
-    creates: "/etc/systemd/system/Splunkd.service"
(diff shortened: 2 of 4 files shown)


SkyKnightSKS:
stop_splunk.yml should be adapted too then

SkyKnightSKS:
I ran this change in a test:
FAILED! => {"changed": false, "msg": "Could not find the requested service splunk: host"}

When I use the service name: 'Splunkd' then it works.

Before systemctl the service was called splunk, after it became Splunkd

SkyKnightSKS:
Basically the fact 'splunk_current_version' can be used to determine the behaviour we need to trigger.
The play should be adapted to use this fact to see if it is 7.2.2 >=, then use system with Splunkd
Before it should use system with splunk.

When the install_splunk.yml task is done, it should update splunk_current_version to be the same as splunk_target_version.  This way the behaviour before the install would be the old init.d script, and after the upgrade it would become systemctl.
However, this does imply that during the upgrade the enable boot-start is also ran again, otherwise the old init.d script would be continued to use even after upgrading to the systemctl supported version.

nwang92:
@SkyKnightSKS how are you running your tests? Within the context of a VM?

Per https://docs.splunk.com/Documentation/Splunk/7.2.3/Admin/RunSplunkassystemdservice#Configure_systemd_using_enable_boot-start, it seems like when `enable boot-start` detects that systemd is available, it will name the service `Splunkd.service`. However, when using sysvinit, it will be referenced as simply `splunk`. This is also configurable with the `-systemd-unit-file-name` flag.

For simplicity, I'd prefer one naming convention such as `splunk`. If we can standardize on that name, it seems like using Ansible's `service` module would handle the implementation across different systems.

SkyKnightSKS:
This particular test was on a debian vm.
Simple is good, the flag + using service module covers all the bases as far as I can tell.

rcabral-splunk:
> Basically the fact 'splunk_current_version' can be used to determine the behaviour we need to trigger.
> The play should be adapted to use this fact to see if it is 7.2.2 >=, then use system with Splunkd
> Before it should use system with splunk.
> 
> When the install_splunk.yml task is done, it should update splunk_current_version to be the same as splunk_target_version. This way the behaviour before the install would be the old init.d script, and after the upgrade it would become systemctl.
> However, this does imply that during the upgrade the enable boot-start is also ran again, otherwise the old init.d script would still be used even after upgrading to the systemctl supported version.
> And the distro needs to have systemctl.

I've originally considered and tested using `splunk_current_version` and `splunk_target_version` facts, however testing specific versions of Splunk isn't the ideal, from the maintenance and config management standpoints. Also, those facts aren't always set, depending on Splunk's product (Enterprise, Insights for Infrastructure, IoT, etc.), due to contents of the manifests files. This is something to be covered in another issue, though.
The solution I've implemented in [4b1ff57](https://github.com/splunk/splunk-ansible/commit/4b1ff57e7419d6a6cf91087ea01a35be9eec2f00) does [manually configures systemd to run splunkd](http://docs.splunk.com/Documentation/Splunk/7.2.3/Workloads/Configuresystemd#Configure_systemd_manually).

SkyKnightSKS:
I don't agree that setting systemd manually without version check is the proper way to go.
If you are using an older version of Splunk (aka pre 7.2.0) systemd is actually not supported.

Or is pre7.2.0 assumed not to be used?

rcabral-splunk:
> I don't agree that setting systemd manually without version check is the proper way to go.
> If you are using an older version of Splunk (aka pre 7.2.0) systemd is actually not supported.
> 
> Or is pre7.2.0 assumed not to be used?

Other products may still use 7.2.0 (which is one my current scenarios) as the base Enterprise version, thus systemd isn't supported, like you well mentioned. For those cases, as well as other Linux platforms like Amazon Linux v1, systemd isn't supported either, no matter if you're trying to run >=7.2.2 . For all these cases, we'd rather to test what's running as PID 1 so we can run splunkd as service accordingly.

SkyKnightSKS:
I fully agree with the method used to detect systemd.  A very clean way of checking support for systemd.

I have read the plays.  It seems that this way it will create both a splunk.service and later on trigger the enable boot-start, which in turn will create a Splunkd.service.

In addition what I meant with pre7.2.0 is, in case the installer is a 7.1 or lower it will still install a splunk.service file and configure the systemd to start that. Which will fail because they don't support _internal_launch_under_systemd and splunk will not start because splunk start will be replaced by the service module.

PR #109, opened 2019-02-11
Bugfix/bundle push wait for shc ready

jonathan-vega-splunk:


ansible_commands/shc_ready.py
```diff
@@ -0,0 +1,66 @@
...
+        self.shc_peers = module.params["shc_peers"]
+        self.user = module.params["spl_user"]
+        self.password = module.params["spl_pass"]
+        self.retry_times = module.params["retry_times"]
+
+    def run(self):
+        URL = "https://{0}:8089/services/shcluster/status?output_mode=json".format(self.captain_url)
+        peers_online = False
+        peers_ready = False 
+        time.sleep(60) # give the SHC some initial time to setup
```
bb03: Do we want to make these sleeps configurable?  Might be something worth tweaking

inventory/environ.py
```diff
@@ -95,6 +95,7 @@ def getDefaultVars():
     defaultVars["splunk"]["deployer_included"] = True if os.environ.get('SPLUNK_DEPLOYER_URL', False) else False
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
+    defaultVars["splunk"]["search_head_cluster_url"] = os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', None)
```
lephino: shouldn't this default to [sh1]:8089/path?

ansible_commands/shc_ready.py
jmervine: I don't think I'd start with a sleep, but rather attempt the connection first and if it fails go in to a sleep loop. There are going to be cases where the SHC is already up and this is an unnecessary delay.

ansible_commands/shc_ready.py
```diff
@@ -0,0 +1,66 @@
...
+        online_peers = None 
+        for _ in range(0, self.retry_times):
+            time.sleep(30) # wait some time before executing next loop
```
jmervine: I'd move this sleep to the end of the look and make it configurable.

ansible_commands/shc_ready.py
```diff
@@ -0,0 +1,66 @@
...
+        self.password = module.params["spl_pass"]
+        self.retry_times = module.params["retry_times"]
+
+    def run(self):
+        URL = "https://{0}:8089/services/shcluster/status?output_mode=json".format(self.captain_url)
+        peers_online = False
+        peers_ready = False 
+        time.sleep(60) # give the SHC some initial time to setup
+        online_peers = None 
+        for _ in range(0, self.retry_times):
```
jmervine: This should optionally be, just wait, no `retry_times`.

inventory/environ.py
jmervine: Would this be clearer as `defaultVars["splunk"]["search_head_captain_url"]`?

roles/splunk_deployer/tasks/main.yml
```diff
@@ -21,6 +21,18 @@
 - name: Flush restart handlers
   meta: flush_handlers
 
+- name: Wait for SHC to be ready 
+  shc_ready:
+    captain_url="{{ splunk.search_head_cluster_url }}"
+    shc_peers="{{ groups['splunk_search_head'] }}"
```
jmervine: The current requirement for having to specify the search head peers is a containerization anti-pattern, and really annoying as an operator. Ideally we could simply boot new search heads and have it self register with the cluster without every having to touch the captain. I realize that's out of scope for this PR though, but it does pose the question, should this be avoided as a requirement in hope of improving things in the future?

Correct me if I'm wrong, but we should be able to push the bundle without the cluster being fully bootstrapped, as long as the captain is up and ready. Then the captain can worry about replication to it's peers. Unless I'm missing something?

ansible_commands/shc_ready.py
```diff
@@ -0,0 +1,66 @@
...
+            spl_pass=dict(required=True, type='str'),
+            retry_times=dict(required=True, type='int')
+        )
+    )
+    shc_ready = ShcReady(module).run()
+    res = dict(changed=False, ansible_facts=shc_ready)
+    module.exit_json(**res)
+
+if __name__ == '__main__':
+    main()
```
jmervine: Not super familiar with Python patterns, but it seems like there should be a new line at the end of the file.

inventory/environ.py
jonathan-vega-splunk: According to the Docs the env variable SPLUNK_SEARCH_HEAD_CAPTAIN_URL is just the hostname of the search head so we have to add the 8089/path.

roles/splunk_deployer/tasks/main.yml
lephino: Totally correct on the pattern, this exists so splunk-ansible can also support running outside of containers as well.  Long term we do plan on supporting compose "scale" commands with a single node, and k8 scale of a service.  That coding logic will likely be dealt with more on the container side than this implementation in splunk-ansible.

jmervine: That’s totally fair and something I didn’t consider. A suggestion would be to support an empty value for search heads and indexers, in conjunction with shc being set to true to trigger logic that causes non-captains to auto join the cluster (without dropping the rep factor 1, as currently does). Again though, out of scope.

ansible_commands/shc_ready.py
```diff
@@ -19,16 +19,16 @@ def __init__(self, module):
...
     def run(self):
         URL = "https://{0}:8089/services/shcluster/status?output_mode=json".format(self.captain_url)
         peers_online = False
         peers_ready = False 
-        time.sleep(60) # give the SHC some initial time to setup
         online_peers = None 
         for _ in range(0, self.retry_times):
-            time.sleep(30) # wait some time before executing next loop
             try:
+                time.sleep(self.shc_bootstrap_delay) # wait some time before executing next loop
```
jmervine: Note: This is still at the top of the loop, I'd put the sleep right after the exception handler (currently after line 45).

inventory/environ.py
arctan5x: where are you adding 8089 to the path?

roles/splunk_deployer/tasks/main.yml
arctan5x: Regarding the Ansible task pattern, I would much prefer to have module just check once when it is executed and use repeat on the task level. This bring more clarity and control over how many times we want to repeat and wait checking on the shc.

inventory/environ.py
jmervine: The current pattern though-out is only and the hostname, this shouldn't break that pattern unless there is a clear and sensible default, rendering this optional.

To @lephino's comment, I'd say it should be
```
defaultVars["splunk"]["search_head_cluster_url"] = os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', deaultVars["splunk"]["search_head_cluster"] + ":8089/path")
```
or something of the like...

PR #113, opened 2019-02-13
Task/circleci

nwang92:
Lots of things done here:
* Added CircleCI integration with splunk-ansible
* Added linting as part of builds
* Adding basic small tests to cover some python scripts within this repo
* Adding large tests (using molecule framework) for end-to-end playbook validation. In general, I found molecule to be a pretty heavy framework (need to drop multiple, duplicate files everywhere to support multiple testing scenarios, etc.) but in the end, I'd rather have a level-of-defense within the splunk-ansible codebase to catch regressions
* Gradual improvement/overhaul of our docs


nwang92:
There are still some sections populated with "TODO", but I think I'd rather make smaller PRs for those instead of enormous changes. As is, I think the structure and majority of the content is fleshed out.

PR #115, opened 2019-02-14
Supporting both init and systemd services in Linux

rcabral-splunk:
systemd is available in CLI's boot-start command as of v7.2.2 only. We need to be able to support both systemd and init to manage splunkd in Linux.


SkyKnightSKS:
The play would need adaptations for Universal Forwarder Support.

roles/splunk_common/files/Splunkd.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+After=network.target
+
+[Service]
+Type=simple
+Restart=always
+ExecStart=/opt/splunk/bin/splunk _internal_launch_under_systemd
```
nwang92: Let's turn this into a Jinja template so we can accodomate variables + support for UF

nwang92:
Can you resolve the merge conflicts here?

roles/splunk_common/files/Splunkd.service
rcabral-splunk: In a separate PR perhaps?

PR #116, opened 2019-02-15
Updates to enable changes made in the docker-splunk/multistage-build branch

mikedickey:
Change default for build_location to be undefined. You can still set this value
to trigger an upgrade etc. Note that the old default value would never work with
existing playbooks because they expect that it has a specific filename format.

Fixed a bug with the version comparison logic when splunk.build_location is set
where it expected version numbers to always be single digit. This would break
for example for maintenance releases greater than '9', ie "7.2.10"

Added an update_etc.yml playbook which is used to run the /sbin/updateetc.sh
script included in the image. This is used to pour files from /opt/splunk-etc
over top /opt/splunk/etc, when an upgrade is necessary (if splunk.version differs).

Updated certain playbooks to only be executed on first time run (ftr)

Fixed a few cases where "splunk.<something> is defined" never works as expected


roles/splunk_common/tasks/get_facts.yml
```diff
@@ -19,23 +19,26 @@
...
+    regexp: 'splunk(?:forwarder)?-(\d+)\.(\d+)\.(\d+)-'
+  when: "'build_location' in splunk and splunk.build_location is not none"
 
 - name: "Find manifests"
   find:
     paths: "{{ splunk.home }}"
     patterns: ".*-manifest$"
     use_regex: yes
   register: manifests
+  when: "'build_location' in splunk and splunk.build_location is not none"
```
SkyKnightSKS: What is the idea to not check for the installed version, based on the fact an installer is defined or not?

mikedickey: This is just an optimization, since `manifests` is only used below, for comparison to the version included in the build installer's filename.

SkyKnightSKS: It is indeed used to do the comparison. But also to detect what the current version is.
This could be used later on to change certain behaviour based on versions if this would later on become necessary and this should be independent if it is freshly installed or not.

mikedickey: Makes sense. I'll remove the `when`s so that `splunk_current_version` gets set.

mikedickey:
> probably want build_location: "/tmp/splunk.tgz" set in environ.py if it's not there? Or does this break backwards compat and requires the other branch to work?

I'm not familiar enough with that script. The intention of this change is to make the default empty, which will be the case with the new images. If it's specified, it should behave the same as before using that install file.

PR #119, opened 2019-02-21
Fixing shell used for pipefail

nwang92:
I broke something with recent linting fixes - it looks like the default shell (/bin/sh) used by the `shell` module doesn't support the `set -o` option. Changing it to explicitly used bash instead. 

Solution provided by https://github.com/ansible/ansible/issues/12903, although I guess it's more of a "hopefully Ansible will fix this one day" type deal.

### DIFF
--- a/roles/splunk_common/tasks/install_apps.yml
+++ b/roles/splunk_common/tasks/install_apps.yml
@@ -44,6 +44,8 @@
 
 - name: Check app contents
   shell: "set -o pipefail && tar --exclude='*/*/*' --exclude='*.*' -tf /tmp/app.spl | awk -F'/' '{ print$1 }' | uniq"
+  args:
+    executable: /bin/bash
   register: app_contents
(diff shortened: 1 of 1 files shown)


roles/splunk_common/tasks/install_apps.yml
```diff
@@ -44,6 +44,8 @@
 
 - name: Check app contents
   shell: "set -o pipefail && tar --exclude='*/*/*' --exclude='*.*' -tf /tmp/app.spl | awk -F'/' '{ print$1 }' | uniq"
+  args:
+    executable: /bin/bash
```
bb03: We might have some complaints about this by people who want to use other shells.

nwang92: Possibly, I would assume bash is available on our Docker images (because we install it) and most AWS AMIs. I'm not sure where else these Ansible plays are being used?

Otherwise, we could void the `set -o pipefail` and find ways to disable the linting.

PR #129, opened 2019-03-04
Feature/add dedicated depserver

nwang92:
Tests to be added in docker-splunk to accomodate for a new role

### DIFF
--- a/docs/ADVANCED.md
+++ b/docs/ADVANCED.md
@@ -1,8 +1,9 @@
 ## Navigation
 
 * [Defaults](#defaults)
-    * [Creating Defaults](#creating-defaults)
-    * [Creating Defaults](#creating-defaults)
+    * [Loading defaults through file](#loading-defaults-through-file)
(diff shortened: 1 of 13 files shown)


nwang92:
There's going to be a lot of docs on this role that will be featured on the docker-splunk side, mostly because this is a pretty powerful role and you could shoot yourself in the foot if done incorrectly.

Outstanding issues from a discussion @bb03 and I had:
In theory, the deployment server could be used against (unclustered) indexers, (unclustered) search heads, standalones, heavy+universal forwarders, et al. But in practice, it is mostly used for the forwarders.

Would we want to still support all the roles above? Or should we limit it to exclude indexers/search heads? The reason being is indexers + search heads have other dedicated, production-grade means of managing app + configuration distribution than the deployment server. 

I guess the question here is: the documentation on deployment server does support the majority of roles, but is that functionality something we would want customers to run in the wild? Should we step in to possibly prevent some nasty topologies?

PR #132, opened 2019-03-06
ensure defaults dir exists before pulling pre/post tasks

rsmitty:
When using the docker-splunk image and trying to pull down post-tasks, I encountered the issue where `/tmp/defaults` doesn't actually exist on the system so ansible fails to pull. This PR adds a step to ensure the configured directory is present.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -12,6 +12,13 @@
       when:
         - ansible_hostname == lookup('env', 'SPLUNK_SEARCH_HEAD_CAPTAIN_URL') or splunk.role == "splunk_search_head_captain"
 
+    - name: Create defaults dir
+      file:
+        path: "{{config.defaults_dir}}"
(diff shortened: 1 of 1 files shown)


site.yml
```diff
@@ -12,6 +12,11 @@
       when:
         - ansible_hostname == lookup('env', 'SPLUNK_SEARCH_HEAD_CAPTAIN_URL') or splunk.role == "splunk_search_head_captain"
 
+    - name: Create defaults dir
+      file:
+        path: "{{config.defaults_dir}}"
+        state: directory
+
```
lephino: probably should set the owner to ansible forcibly, just to be safe.

rsmitty: Right on, that makes sense. Amended my commit to add that in.

PR #137, opened 2019-03-07
Task/playbook consolidation

nwang92:
I'm going to make some tests, I realized we didn't have test coverage over pre/post playbooks


execute_adhoc_plays.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Create defaults dir
```
arctan5x: would you like to add escalation to root here to prevent any permission issues?

nwang92: This part is a bit tricky and needs some further testing. 

When ansible runs, it assumes the user `splunk`. With this playbook as-is, this directory is created and owned by the `splunk` user. All the subsequent downloaded playbooks will be written into this `/tmp/defaults` under `splunk`.

As long as our ansible process runs everything as `splunk`, there shouldn't need to be any permission changing that happens. However, I'm not certain what would happen in the case of privilege-escalation within a custom playbook, so that part needs some validation.

arctan5x: I think as long as we escalate, we can create a folder anywhere in the file system. Then, we make it owned by a splunk user so that splunk user can further use it. I guess my worry is that is there ever a need for a user to create a defaults directory somewhere that splunk user doesn't have permission? tmp is open to everyone, but just curious if there are other cases.

lephino: why not just set this play to be owned by "ansible" and have a 775 right?  Then it handles both use cases, and makes sure our ansible user can reference it?

nwang92: I'm a little hesitant to call out a `user: ansible` because at the moment, we don't do this anywhere. Adding that will expand our user/group footprint with an additional dependency for such a minimal feature.

lephino: Then why not just say at the start of the play "check if $defaults_home folder exists, if not, create under the current user".  I think this is fine to run as a conditional and then open up the dir.

nwang92: The `file`/`state: directory` module should already take care of this - it will only create the directory if it doesn't exist. It also current does create it under the current user, but since our ansible.cfg comes baked with:
```
[privilege_escalation]
become = Yes
become_method = sudo
become_user = splunk
```
It does mean this directory will belong to the splunk user, ex:
```
ansible@0674b5242828:/tmp$ ls -ltrah
total 327M
-rw-r--r-- 1 ansible ansible 327M Feb  7 20:26 splunk-7.2.4-8a94541dcfac-Linux-x86_64.tgz
drwxr-xr-x 1 root    root    4.0K Mar  8 01:04 ..
drwxr-xr-x 2 splunk  splunk  4.0K Mar  8 01:05 defaults
```

nwang92: Hmm we may want to changes the permissions slightly to allow `o+w` but that may be the only modification I see?

PR #143, opened 2019-03-14
Multisite

bb03:
First pass at multisite plays.  Please see the docker-splunk multisite branch for a scenario.  If you want to test it locally, please check out the changes to that branch

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -96,7 +96,7 @@ def getDefaultVars():
     defaultVars["splunk"]["deployer_included"] = True if os.environ.get('SPLUNK_DEPLOYER_URL', False) else False
     defaultVars["splunk"]["indexer_cluster"] = True if os.environ.get('SPLUNK_CLUSTER_MASTER_URL', False) else False
     defaultVars["splunk"]["search_head_cluster"] = True if os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', False) else False
-    defaultVars["splunk"]["search_head_cluster_url"] = os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', None) 
+    defaultVars["splunk"]["search_head_cluster_url"] = os.environ.get('SPLUNK_SEARCH_HEAD_CAPTAIN_URL', None)
     # Need to provide some file value (does not have to exist). The task will automatically skip over if the file is not found. Otherwise, will throw an error if no file is specified.
     defaultVars["splunk"]["license_uri"] = os.environ.get('SPLUNK_LICENSE_URI', 'splunk.lic')
     if defaultVars["splunk"]["license_uri"] and '*' in defaultVars["splunk"]["license_uri"]:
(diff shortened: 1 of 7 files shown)


roles/splunk_cluster_master/tasks/initialize_multisite_master.yml
```diff
@@ -0,0 +1,22 @@
+---
+- include_tasks: ../../../roles/splunk_common/tasks/wait_for_splunk_instance.yml
+  vars:
+    splunk_instance_address: "{{ splunk.multisite_master }}"
+
+- name: Convert Extrenal Cluster Master Name into Internal URI
+  set_fact:
+      multisite_master_uri: "https://{{ splunk.multisite_master }}:{{ splunk.multisite_master_port }}"
```
jmeixensperger: Might want to use {{ cert_prefix }} here instead of "https" because of my recent changes

inventory/environ.py
```diff
@@ -191,6 +198,16 @@ def overrideEnvironmentVars(vars_scope):
     vars_scope["splunk"]["http_enableSSL_cert"] = os.environ.get('SPLUNK_HTTP_ENABLESSL_CERT', vars_scope["splunk"]["http_enableSSL_cert"])
     vars_scope["splunk"]["http_enableSSL_privKey"] = os.environ.get('SPLUNK_HTTP_ENABLESSL_PRIVKEY', vars_scope["splunk"]["http_enableSSL_privKey"])
     vars_scope["splunk"]["http_enableSSL_privKey_password"] = os.environ.get('SPLUNK_HTTP_ENABLESSL_PRIVKEY_PASSWORD', vars_scope["splunk"]["http_enableSSL_privKey_password"])
+    #Used for multisite
+    if 'SPLUNK_SITE' in os.environ or 'site' in vars_scope["splunk"]:
+        vars_scope["splunk"]["site"] = os.environ.get('SPLUNK_SITE', vars_scope["splunk"].get("site"))
+        vars_scope["splunk"]["all_sites"] = os.environ.get('SPLUNK_ALL_SITES', vars_scope["splunk"].get("all_sites"))
+        vars_scope["splunk"]["multisite_master"] = os.environ.get('SPLUNK_MULTISITE_MASTER', vars_scope["splunk"].get("multisite_master"))
+        vars_scope["splunk"]["multisite_master_port"] = os.environ.get('SPLUNK_MULTISITE_MASTER_PORT', vars_scope["splunk"].get("multisite_master_port", 8089))
```
nwang92: Wondering if we need this new env var - in general, we always assume `splunk.svc_port` for management port.

nwang92: Actually, I think it would be helpful to expose a way to connect to remote Splunk instances on non-traditional ports, but I think we should talk about how that would be done? I like the idea of supporting `SPLUNK_MULTISITE_MASTER=cm1` or `SPLUNK_MULTISITE_MASTER=cm1:8089` versus adding yet another environment variable :P 

This could also extend to connecting ephemeral containerized search heads to long-running indexers somewhere else (like EC2) via something like `SPLUNK_INDEXER_URL=idx1:8089,idx2:8090,idx3:18089`

roles/splunk_cluster_master/tasks/initialize_multisite_master.yml
bb03: Thanks, I'll do that

inventory/environ.py
bb03: I'm fine changing it back to svc_port.  I was thinking about non-traditional ports, but its probably better that until we get a request for it, we shouldn't add another variable.

PR #145, opened 2019-03-14
Release 7.2.5

arctan5x:


version.txt
```diff
@@ -1 +1 @@
-os0.2.0
+7.2.5
```
nwang92: This file actually gets modified in the `make ansible` command from docker-splunk, but it's good to update this too :P

PR #150, opened 2019-03-25
Include a default cert_prefix

bb03:
@jmeixensperger Make sure that this fits in with your prefix changes.  It should just set a default based on an environment variable, but I want to make sure its compatible with your other stuff.


bb03:
@jmeixensperger Thanks, I'm okay with this getting overwritten.  Its more about making sure that we set a default value.  I'll take a look at the multisite plays and see about including your cert prefix play, which would be more accurate.

PR #163, opened 2019-04-16
Fixing play names for pre/post playbooks

nwang92:
There's some interesting behavior with `include_tasks`. From my testing, it seems like when Ansible loads the tasks dynamically from this module, it seems to cache/store the tasks based on filename. So it's possible to have a pre-playbook named `/tmp/defaults/playbook.yml` that gets included, executed, then immediately deleted. But if the contents of `/tmp/defaults/playbook.yml` changes (in the case of post-playbook), the next subsequent run of `include_tasks: /tmp/defaults/playbook.yml` actually runs the former (deleted) plays.

Changing this here so that when we do save playbooks, we use the user-provided filenames. There is still a corner-case in which this will break, which is when a user provides multiple files with the same name (but different contents), ex. `SPLUNK_ANSIBLE_PRE_TASKS=file:///example.yml` and `SPLUNK_ANSIBLE_POST_TASKS=file:///example.yml`

### DIFF
--- a/execute_adhoc_plays.yml
+++ b/execute_adhoc_plays.yml
@@ -7,7 +7,8 @@
 - name: Fetch adhoc playbooks
   get_url:
     url: "{{ playbook }}"
-    dest: "{{ config.defaults_dir + '/playbook.yml' }}"
+    dest: "{{ config.defaults_dir '/' + playbook|basename }}"
+    force: yes
   ignore_errors: yes
   register: downloaded_plays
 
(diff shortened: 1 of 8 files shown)


execute_adhoc_plays.yml
```diff
@@ -7,7 +7,8 @@
 - name: Fetch adhoc playbooks
   get_url:
     url: "{{ playbook }}"
-    dest: "{{ config.defaults_dir + '/playbook.yml' }}"
+    dest: "{{ config.defaults_dir '/' + playbook|basename }}"
```
jmervine: Nice solution! You're missing a `+` after `config.defaults_dir`, should be
```
dest: "{{ config.defaults_dir + '/' + playbook|basename }}"
```

I was going to suggest something like
```
dest: "{{ config.defaults_dir + '/' + playbook|hash('sha1') + '-playbook.yml' }}"
```

... but your solution is probably better.

PR #166, opened 2019-04-30
Use systemd module parameter called daemon-reload.

akulbe:
The proper way to do this according to the Ansible docs for the systemd module is to use the parameter called daemon-reload. We encountered this when we were using your repo here, locally.

Here's the reference for it: https://docs.ansible.com/ansible/latest/modules/systemd_module.html?highlight=systemd

Thank you for your work!

### DIFF
--- a/roles/splunk_common/tasks/enable_service.yml
+++ b/roles/splunk_common/tasks/enable_service.yml
@@ -22,8 +22,9 @@
   become: yes
   become_user: "root"
   systemd:
-      name: daemon-reload
-      state: restarted
+      daemon-reload: yes
+      name: splunk.service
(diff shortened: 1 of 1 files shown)


roles/splunk_common/tasks/enable_service.yml
```diff
@@ -22,8 +22,9 @@
   become: yes
   become_user: "root"
   systemd:
-      name: daemon-reload
-      state: restarted
+      daemon-reload: yes
+      name: splunk.service
```
nwang92: Is `name` required here for daemon-reload? We may possibly just want to tell systemd to re-read all configs here?

Issue #140, opened 2019-03-10
http_port value in default.yml ignored.


BertieW:
Setting http_port in default.yml does nothing. The port is always 8000.

arctan5x:
Let me take a look.

BertieW:
We already have something serving on 8000 on that machine. It isn't anything that we can't work around but it is confusing that the ability to specify the port is there and it is a value that can be changed via the SplunkWeb or config files. 

If the script always provisions splunkweb to 8000 at boot, maybe documentation can be amended to explain that?

gibban69:
> We already have something serving on 8000 on that machine. It isn't anything that we can't work around but it is confusing that the ability to specify the port is there and it is a value that can be changed via the SplunkWeb or config files, yet nothing is done with that info.
> 
> If the script always provisions splunkweb to 8000 at boot, maybe documentation can be amended to explain that?

I agree that it is very confusing. I takes time to discover that changing some of the values in the generated default.yml has no effect. There is no indication which ones are just ignored. I suggest the "create-defaults" commands only generates the config parameters that are not ignored.

PR #169, opened 2019-05-03
GitHub issue 140

BertieW:
Resolves https://github.com/splunk/splunk-ansible/issues/140 and https://github.com/splunk/docker-splunk/issues/179. 

Adds a new task which amends web.conf when user-specified http_port in default.yml is not 8000. Molecule test playbook.yml in splunk_standalone role also changed to validate.

### DIFF
--- a/roles/splunk_standalone/molecule/default/playbook.yml
+++ b/roles/splunk_standalone/molecule/default/playbook.yml
@@ -53,7 +53,7 @@
       http_enableSSL_cert: null
       http_enableSSL_privKey: null
       http_enableSSL_privKey_password: null
-      http_port: 8000
+      http_port: 8080
       idxc:
         enable: false
         label: idxc_label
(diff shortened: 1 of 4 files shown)


roles/splunk_standalone/molecule/default/playbook.yml
```diff
@@ -53,7 +53,7 @@
       http_enableSSL_cert: null
       http_enableSSL_privKey: null
       http_enableSSL_privKey_password: null
-      http_port: 8000
+      http_port: 8080
```
bb03: In general this looks fine to me, but I would like a short comment here about why this changed to 8080.  Mostly so that lazy engineers like me don't accidentally revert it.

BertieW: I've changed it to test the custom port scenario. I'm sorry, a bit new at this. Should I add that as a comment in the code and check back in?

bb03: The comment in the code.  What you have right now is perfect. Thanks!

Issue #165, opened 2019-04-17
Add support for root_endpoint in web.conf


d3vil-st:
default.yml missing option for root_endpoint, that could be needed if you deploying splunk in kubernetes behind ingress in subpath URL. For example it is not working with this ingress:
```yaml
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: splunk
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
  - http:
      paths:
      - path: /splunk/
        backend:
          serviceName: master
          servicePort: splunkweb
```
Continuously redirecting from http://server/splunk/ to http://server/en-US/splunk/

c3-rockyong:
you have to put the following in your default.yml:
```
splunk:
  conf:
    web:
      directory: /opt/splunk/etc/system/local/
      content:
        settings:
          root_endpoint: /splunk
```

that will create a custom web.conf that does what you want. took me forever to discover this bit too

bb03:
@arctan5x @nwang92  We should probably make our documentation better around this.

PR #171, opened 2019-05-03
Feature/root endpoint

nwang92:
Adding support for https://github.com/splunk/splunk-ansible/issues/165


jmeixensperger:
Should we update any docs for this?

nwang92:
Yes, docs change would be required. I'll get that in this PR

PR #174, opened 2019-05-07
AWS Windows Support

jmeixensperger:
- Windows platforms now use the 'runas' become_method, as well as the 'Administrator' become_user for privilege escalations. It will have a windows-specific config file that can be copied into the normal 'ansible.cfg' path for use.
- Windows install can perform upgrades, removes the installer file after use, and is able to download remote packages with the 'msiexec' command.


roles/splunk_common/tasks/get_facts.yml
```diff
@@ -55,4 +60,4 @@
 # We are upgrading if it is not a fresh installation and the current version is different from the target version
 - name: "Setting upgrade fact"
   set_fact:
-    splunk_upgrade: "{{ 'build_location' in splunk and splunk.build_location is not none and not splunk_install and splunk_target_version != splunk_current_version | default(False) }}"
+    splunk_upgrade: "{{ not splunk_install and splunk_target_version and splunk_target_version != splunk_current_version | default(False) }}"
```
jmeixensperger: splunk.build_location is always being set now, however splunk_target_version can be an empty string if there's no target build.

PR #178, opened 2019-05-17
regex fix for debug build

jmeixensperger:
Allows debug build names containing "splunk-unstripped-x.x.x" and local builds that may not have the ending hyphen (ex: "splunk-x.x.x.tgz")

### DIFF
--- a/roles/splunk_common/tasks/get_facts.yml
+++ b/roles/splunk_common/tasks/get_facts.yml
@@ -40,7 +40,7 @@
   set_fact:
     splunk_target_version: "{{ splunk.build_location | regex_search(regexp, '\\1') | default('0') }}"
   vars:
-    regexp: 'splunk(?:forwarder|light|cloud|-unstripped)?-((\d+)\.(\d+)\.(\d+))'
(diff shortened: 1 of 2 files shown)


jmeixensperger:
Looks like the CI is failing due to package incompatibilities in molecule. I'm tracking the changes/release and will re-run the job when it's in.

roles/splunk_common/tasks/get_facts.yml
```diff
@@ -40,7 +40,7 @@
   set_fact:
     splunk_target_version: "{{ splunk.build_location | regex_search(regexp, '\\1') | default('0') }}"
   vars:
-    regexp: 'splunk(?:forwarder|light|cloud)?-((\d+)\.(\d+)\.(\d+))-'
+    regexp: 'splunk(?:forwarder|light|cloud|-unstripped)?-((\d+)\.(\d+)\.(\d+))'
```
bb03: You only care about the last digits with this, right?  (i.e. 7.2.5)
So why not just use this ```splunk\D*?-((\d+)\.(\d+)\.(\d+))$```

PR #193, opened 2019-06-12
Fixing parameter used by molecule playbooks

nwang92:
Fix for https://github.com/splunk/splunk-ansible/pull/191#issuecomment-500632289

Looks like there's some bad copy-pasta when these vars were written into the molecule playbook. Previously, this was being treated as `splunk.splunk_home_ownership_enforcement` and instead it should've been simply `splunk_home_ownership_enforcement`.


nwang92:
Looks like there are subsequent idempotency failures ~ let me address those.
```
* [instance] => splunk_common : Update /opt/splunk/etc
```

PR #197, opened 2019-06-17
Address synchronization issues

bb03:
So, a little background.  During some internal testing we saw issues where a search head cluster would appear to desync with some of its members (this happened during the S3 (search head cluster, no index cluster) SVA setup.  I happened to see it too, although rarely, and this is the fix for it.  However, it does come with some negatives.  First of all, it increases the overall deployment time of the S3 scenario by over 100 seconds (taking it from ~114 seconds to ~218 for me), so I am worried about how this performs at a massive scale.  I may try to loop additional people in before submitting this as a fix, because of the performance implications . This is what the code is based on https://docs.splunk.com/Documentation/Splunk/7.3.0/DistSearch/HowconfrepoworksinSHC - under the "Perform a manual resync" section.

Its a good question if this belongs here or if this is something that needs to get pushed onto the consumers (here's how to resync if you see this error, etc etc.) I'm sort of in favor of the latter, but willing to entertain different options.

### DIFF
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -69,3 +69,19 @@
   notify:
     - Restart the splunkd service
   no_log: "{{ hide_password }}"
+
+- include_tasks: ../../../roles/splunk_common/tasks/wait_for_splunk_instance.yml
+  vars:
+    splunk_instance_address: "{{ groups['splunk_deployer'][0] }}"
+
+- name: Destructive sync search head
+  command: "{{ splunk.exec }} resync shcluster-replicated-config -auth {{ splunk.admin_user }}:{{ splunk.password }}"
+  become: yes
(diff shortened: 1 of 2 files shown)


jmeixensperger:
I agree that this is something we should provide to consumers on support cases. I don't think we should slow down the scenario deployment time unless this becomes a more widespread/frequent issue.

nwang92:
Do you think this is something this task/script should be addressing?  https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_deployer/tasks/main.yml#L24
https://github.com/splunk/splunk-ansible/blob/develop/ansible_commands/shc_ready.py

I'm wondering if there's gaps in that readiness script that we can enhance to prevent the need for a manual resync?

bb03:
Thats something I was investigating.  I had intermittent results about this message going away.  Sometimes it went away after 2-3 minutes, sometimes it stayed indefinitely (well beyond 10 minutes).

PR #198, opened 2019-06-18
Fixing unresolved vars

nwang92:
I think the last PR's use of the nested vars in the task name broke things. I believe this is due to the fact that the variable is defined at the inventory-scope and not globally to Ansible. For instance, there's some undefined behavior if this were run against 2 different hosts with 2 different settings.


nwang92:
Also, I think the way molecule tests are structured, we didn't catch this.

PR #207, opened 2019-06-27
Windows java

arctan5x:


roles/splunk_common/tasks/java_tasks/install_openjdk9_jdk_windows.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Download the Java distribution
+  get_url:
+    url: "https://download.java.net/openjdk/jdk9/ri/jdk-9+181_windows-x64_ri.zip"
+    dest: /home/splunk/
```
nwang92: Is there a way we can make this hardcoded path modular?

jrigassio-splunk: no. The jdk9 is probably done with development and it would be stable to use a fixed version. Even if we make it modular, someone needs to look up the version and supply as a parameter. Not worth $$$

jrigassio-splunk: - tony

PR #215, opened 2019-07-16
Bugfix/infra 11591

jrigassio-splunk:


library/es_ta_for_indexers.py
```diff
@@ -23,13 +24,34 @@ def create_parser():
...
+        spec.update()
+        archive = makeIndexTimeProperties(spec, session_key)
+        success = True
+    except TypeError:
+        #Some versions have a change that consolidated app_info, namespace, and include_indexes, 
+        #and added include_properties.
+        #Below code is written to handle older versions.
+        pass
+    if success:
+        print archive
```
arctan5x: print

jrigassio-splunk: https://github.com/splunk/splunk-ansible/blob/develop/library/es_ta_for_indexers.py#L41

jrigassio-splunk: My understanding is that the function is meant to print upon successfully creating the archive? Unless the older one snuck in as well.

PR #217, opened 2019-07-18
Fixing regex to support patch versioning

nwang92:


nwang92:
Found the former regex was a problem for 7.2.5.1

nwang92:
Fixed some false error messaging issues as seen below:
```
TASK [splunk_common : Get Splunk status] ***************************************
    fatal: [instance]: FAILED! => {"changed": false, "cmd": ["/opt/splunk/bin/splunk", "status", "--accept-license", "--answer-yes", "--no-prompt"], "delta": "0:00:00.408664", "end": "2019-07-18 21:55:14.623227", "msg": "non-zero return code", "rc": 3, "start": "2019-07-18 21:55:14.214563", "stderr": "", "stderr_lines": [], "stdout": "splunkd is not running.", "stdout_lines": ["splunkd is not running."]}
    ...ignoring
```
I wanted to suppress the fatal message above because although we ignore it, end users might see this as a potential error when it's actually not a problem.

I also fixed an error that's causing a test failure on docker-splunk: https://circleci.com/gh/splunk/docker-splunk/353?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

roles/splunk_standalone/molecule/default/playbook.yml
```diff
@@ -27,7 +27,7 @@
       license_uri: null
       wildcard_license: false
       build_remote_src: true
-      build_location: ${SPLUNK_INSTALLER}
+      build_location: https://download.splunk.com/products/splunk/releases/7.3.0/linux/splunk-7.3.0-657388c7a488-Linux-x86_64.tgz
```
jmeixensperger: What's the reasoning here? Won't we have to manually update this now?

nwang92: This is just used for the molecule tests. The previous setting of `${SPLUNK_INSTALLER}` actually doesn't work anymore - I'm guessing that's an artifact of our old image pattern? I couldn't find any references to a default env var `SPLUNK_INSTALLER` in docker-splunk.

nwang92: And no, we shouldn't need to update this unless the way we build the image changes again - this is just used for tests so it should dynamically fetch the 7.3.0 build and setup the container using that.

PR #218, opened 2019-07-18
On an upgrade, make sure to remove any residual python files

bb03:
On an upgrade, some versions of splunk left behind python files we do not want to be there.  Removing these prior to an upgrade.

### DIFF
--- a/roles/splunk_common/tasks/install_splunk.yml
+++ b/roles/splunk_common/tasks/install_splunk.yml
@@ -10,6 +10,17 @@
   - "{{ manifests.files }}"
   when: splunk_upgrade | bool
 
+- name: Remove old python files
+  file:
+    path: "{{ item.path }}"
+    state: "absent"
+  ignore_errors: yes
+  become: yes
+  become_user: "{{ privileged_user }}"
+  with_items:
+  - "{{ py_files.files }}"
+  when: splunk_upgrade | bool
+
 - name: Install Splunk (Linux)
   unarchive:
     src: "{{ splunk.build_location }}"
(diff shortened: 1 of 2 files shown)


roles/splunk_common/tasks/install_splunk.yml
```diff
@@ -10,6 +10,17 @@
   - "{{ manifests.files }}"
   when: splunk_upgrade | bool
 
+- name: Remove old python files
```
nwang92: LGTM - only comment would be that we might be able to merge this with the "Remove old manifest files" task above?

bb03: I was talking with @lephino and we think we can use the old manifest to dictate what files to remove. This is likely a slightly larger change though so I would like to merge this in, so that we have a couple days and can resolve this issue from blocking others.

roles/splunk_common/tasks/install_splunk.yml
```diff
@@ -10,6 +10,17 @@
...
+  file:
+    path: "{{ item.path }}"
+    state: "absent"
+  ignore_errors: yes
+  become: yes
+  become_user: "{{ privileged_user }}"
+  with_items:
+  - "{{ py_files.files }}"
```
jmeixensperger: Can we use the paths here? It would eliminate the need for searching them, and it already ignores when we try to remove paths that don't exist.

bb03: I think I might do something even more involved tomorrow.  We'll probably use the manifests file to determine what files should be deleted and remove anything that matches the hash in there.

PR #233, opened 2019-08-04
Ansible should use FQDN of host when initializing SHC

mikedickey:
As per Splunk docs (https://docs.splunk.com/Documentation/Splunk/7.3.0/DistSearch/SHCdeploymentoverview#4._Initialize_cluster_members), the "splunk init shcluster-config" command should use fully qualified names:

The -mgmt_uri parameter specifies the URI and management port for this instance. You must use the fully qualified domain name. This parameter is required.

### DIFF
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -17,7 +17,7 @@
     splunk_instance_address: "{{ groups['splunk_deployer'][0] }}"
 
 - name: Initialize SHC cluster config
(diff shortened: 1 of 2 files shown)


mikedickey:
I ran full tests on this repo + docker-splunk and all tests passed with this change.

lephino:
> my fear on this change was always that if you used a volume, and then mounted a newer container, your FQDN would change.

I spoke with Mike offline about my concerns, and think they can be overcome by using compose / k8 / --hostname.

zhengyuli:
@mikedickey @bb03 
this PR has broken S3/C3 provision for EC2 Windows
as Mike said the "splunk init shcluster-config" command should use fully qualified names, but also the server list in "splunk bootstrap shcluster-captain -servers_list" should be consistent with server name configured in "splunk init shcluster-config"

bb03:
I took a look at this, and the fix that I'm trying seems to be a little heavier than a quick fix.  For windows environments, it seems like the fqdn is resolving to something like
sh1.region.compute.internal, so it's really not resolvable by even the localhost.  Yay windows.

Going to revert for now until we can get a plan in place that deals with windows too.

PR #237, opened 2019-08-13
Revert "Ansible should use FQDN of host when initializing SHC"

bb03:
Reverts splunk/splunk-ansible#233

Reverting this because it causes an issue with Windows.  It seems like the fqdn on windows is something like "sh1.aws-region.compute.internal" which is unresolvable.

### DIFF
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -17,7 +17,7 @@
     splunk_instance_address: "{{ groups['splunk_deployer'][0] }}"
 
 - name: Initialize SHC cluster config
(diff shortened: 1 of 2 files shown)


mikedickey:
Declining since reverting this would now break other things, and I'm not a fan of kicking the can unless it's truly urgent. Correct default behavior IMHO is to use FQDN; if DNS is broken on Windows I think there should be a flag or something added to work around that.

bb03:
@mikedickey Seems reasonable, may I assign it to you for the fix?

bb03:
Closing and deleting branch, per an offline discussion we are retaining the fix but making other adjustments to accommodate different networking scenarios, especially around windows.

PR #238, opened 2019-08-16
Use splunk user to get splunk facts

mikedickey:
There were certain facts accessible to (and owned by) the splunk user that we previously were upgraded to priviliged user to get. This PR updates behaviour so that the splunk user is used to get these facts.

### DIFF
--- a/roles/splunk_common/tasks/get_facts.yml
+++ b/roles/splunk_common/tasks/get_facts.yml
@@ -9,7 +9,7 @@
   stat:
     path: "{{ splunk.exec }}"
   become: yes
-  become_user: "{{ privileged_user }}"
+  become_user: "{{ splunk.user }}"
   register: pre_existing_splunk_exec
 
 - name: "Set splunk install fact"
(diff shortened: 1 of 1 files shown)


bb03:
@mikedickey Why is this change necessary?  Is it required as a pre-requisite for some other change or does directly address any customer issues?

mikedickey:
@bb03 this is a security fix. Escalation should only be performed to the minimal scope necessary, and in this case it's the splunk user. I presume the original code was just an oversight from copy/pasting, or not thinking too much about potential consequences of using root permissions to check files owned by the splunk user.

Also, it creates problems with my other PRs because it effectively requires that you can only run the ansible playbooks with root privs. This is unnecessary and undesirable for many reasons. After making this change, root privs are only required when necessary, by specific features that actually need them.

bb03:
@mikedickey Unfortunately, your presumption looks like it regresses some behavior.  This change was originally committed specifically in PR https://github.com/splunk/splunk-ansible/pull/211/files and it looks like this was to address some very specific ownership issues.  Can you please verify that with these changes upgrade still works?

PR #240, opened 2019-08-19
fix the print usage in python3

hzhzhang:
fix the print usage in python3


hzhzhang:
@bb03 @nwang92  pls help review.

PR #241, opened 2019-08-19
Add support for overriding splunk.user via SPLUNK_USER environment variable

mikedickey:
Currenty, you can only override the user splunk runs as by using a default.yml file that overrides splunk.user.

For Universal Forwarders especially, it is sometimes desired to run splunkd as root or some other user so that it is able to read log files generated and owned by other processes.

This just makes it possible to change the user splunk runs as using the SPLUNK_USER environment variable, the same was as the SPLUNK_GROUP environment variable can be used to override splunk.group

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -174,6 +174,7 @@ def getSplunkApps(vars_scope):
         vars_scope["splunk"]["apps_location"] = []
 
 def overrideEnvironmentVars(vars_scope):
+    vars_scope["splunk"]["user"] = os.environ.get("SPLUNK_USER", vars_scope["splunk"]["user"])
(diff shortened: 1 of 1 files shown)


vinitmasaun:
Hello,

I am trying to use the SPLUNK_USER and SPLUNK_GROUP env vars to run my splunk uf container in kubernetes as another non-root user but the splunk container goes to crashloop with the following error reported by one of the ansible tasks:

Friday 21 May 2021 03:36:47 +0000 (0:00:00.027) 0:00:01.376 ************

TASK [splunk_common : Check for scloud] ****************************************
fatal: [localhost]: FAILED! => {
"changed": false,
"rc": 1
}

MSG:

MODULE FAILURE
See stdout/stderr for the exact error


MODULE_STDERR:

sudo: unknown user: appusr
sudo: unable to initialize policy plugin

PR #245, opened 2019-08-20
Fixing new import and doing a little linting

bb03:
Essentially, we should use a different location (if present) for make_splunkhome_path


jmeixensperger:
Looks good, currently testing w/ indexer clustering

PR #250, opened 2019-09-03
Feature/use fqdn as hostname

zhengyuli:
as per suggestions from @mikedickey and @bb03 , orca side will pass in SPLUNK_HOSTNAME env to each splunk instance, which will be consumed by splunk_ansible/inventory/environ.py as splunk.hostname to replace ansible_fqdn used anywhere.


bb03:
@zhengyuli Can you make a separate PR to update the docs?

zhengyuli:
@bb03  sure.

PR #265, opened 2019-09-24
Replacing six urllib compatibility with requests

nwang92:
I don't think we need the six urllib backwards compatibility library here, because we're already importing the python requests package.

There's also some unit tests added around this method to catch issues like this from occurring.


nwang92:
This should fix: https://github.com/splunk/docker-splunk/issues/241

PR #267, opened 2019-09-27
Bugfix of s3 provision for windows

zhengyuli:
for S3/C3 deployment failure on windows platform we found the delay time for resync and wait for splunk instance is too short to check, with the delay time adjusted we can decrease failure probability from 30% to 2%.

### DIFF
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -70,7 +70,7 @@
     - Restart the splunkd service
   no_log: "{{ hide_password }}"
 
-- name: Destructive sync search head
+- name: Destructive sync search head (Linux)
   command: "{{ splunk.exec }} resync shcluster-replicated-config -auth {{ splunk.admin_user }}:{{ splunk.password }}"
   become: yes
   become_user: "{{ splunk.user }}"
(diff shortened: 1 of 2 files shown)


roles/splunk_search_head/tasks/search_head_clustering.yml
```diff
@@ -80,4 +80,18 @@
...
+  command: "{{ splunk.exec }} resync shcluster-replicated-config -auth {{ splunk.admin_user }}:{{ splunk.password }}"
+  become: yes
+  become_user: "{{ splunk.user }}"
+  when: not splunk_search_head_captain | bool and splunk.preferred_captaincy | bool
+  register: task_result
+  changed_when: task_result.rc == 0
+  until: task_result.rc == 0
+  retries: "{{ retry_num }}"
+  delay: 60
+  when: ansible_system is match("CYGWIN*|Win32NT")
```
jmeixensperger: The linting is failing due to 2 "when" keys.

nwang92:
Can we just update the `delay` without making 2 separate plays? These seem pretty identical, and either way the linux provisioning plays will still exit earlier while giving more time for the windows.

zhengyuli:
@nwang92 I am worried about the delay time for Linux is too long if merging two tasks together, as delay option will take effect before the first try. Actually there is no side effect to merge these two tasks together except slowing the progress of provision for linux down, but on the other hand it will also decrease the failure probabilities for deploying large-scale deployments simultaneously.
Anyway, I am ok if you are all agreed with this idea.

nwang92:
@zhengyuli I'm ok with a longer delay interval to cover both linux- and windows- based deployments under the same task. I would also drop a brief comment inline to explain why the delay is longer.

zhengyuli:
@nwang92 cool!

PR #274, opened 2019-10-08
allow manually specifying forward-servers

tod-uma:
In our environment, we have a set of heavy forwarders that are between our indexers and our universal forwarders. This pull request allows us to manually specify our forward servers since there isn't a good way to automatically determine which servers are the ones we want everything to go through.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ large-tests: test-setup
 	@echo 'Running the super awesome large tests'
 	cd roles/splunk_standalone && molecule test
 	cd roles/splunk_universal_forwarder && molecule test
+	cd roles/splunk_heavy_forwarder && molecule test --all
 
 py3k-large-tests: py3k-test-setup
 	@echo 'Running the super awesome large tests'
--- a/roles/splunk_common/handlers/restart_splunk.yml
+++ b/roles/splunk_common/handlers/restart_splunk.yml
@@ -11,7 +11,7 @@
 
 - name: "Restart the splunkd service - Via systemd"
   service:
-    name: Splunkd
+    name: "{{ splunk.service_name }}"
     state: restarted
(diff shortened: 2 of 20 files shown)


bb03:
Hi @tod-uma Can you tell me what kind of tests you ran for these changes?  Especially if you have any test artifacts (like a docker compose template) we would really like to incorporate that into our testing.

tod-uma:
> Hi @tod-uma Can you tell me what kind of tests you ran for these changes? Especially if you have any test artifacts (like a docker compose template) we would really like to incorporate that into our testing.

What I do to test is I run it on our dev systems, which are just vmware vms, then when I'm confident it'll work, I move to our production systems. I'm not familiar with docker compose. Can  you point me to some documentation on how to use it? I'd be happy to run other tests on anything I submit.

tod-uma:
> > Hi @tod-uma Can you tell me what kind of tests you ran for these changes? Especially if you have any test artifacts (like a docker compose template) we would really like to incorporate that into our testing.
> 
> What I do to test is I run it on our dev systems, which are just vmware vms, then when I'm confident it'll work, I move to our production systems. I'm not familiar with docker compose. Can you point me to some documentation on how to use it? I'd be happy to run other tests on anything I submit.

ok, I've got docker working on my desktop box and I'm able to get the make test to mostly work. However, I'm running into an issue where it fails to launch splunk due to an "unsupported filesystem". I suspect that is because my desktop runs on zfs. I'll have to look into that more.

bb03:
@tod-uma Okay thanks, we'll do some testing on our side before we merge it in, so be patient with us while we go through our verification.

tod-uma:
So I've been learning molecule to perform tests. I needed to add support for passing a proxy server to the docker containers. I also added tests specific to the manually specifying forward servers. I'm glad I did because I also found bugs while adding the tests.

PR #276, opened 2019-10-10
fix bug that local package can't be installed correctly

mrssss:
The file protocal is not supported by unarchive module. If the src option is a local path, this playbook can't work.


roles/splunk_common/tasks/install_splunk_tgz.yml
```diff
@@ -1,14 +1,29 @@
...
   register: install_result
   until: install_result is succeeded
   retries: "{{ retry_num }}"
   delay: 3
   become: yes
   become_user: "{{ privileged_user }}"
 
+- name: Install Splunk (Linux) from local build
+  unarchive:
+    src: "{{ '/opt/container_artifact/' + splunk.build_location|basename }}"
```
bb03: I have some issues with this fix as it stands now.  With this string hardcoded, its assumed that this directory will always exist, when that may not be the case depending on the platform and the environment.

mrssss: fixed hard code issue, and using regex filter to remove 'file://'

PR #282, opened 2019-10-25
Add support for cluster labels via env vars.

jmervine:
It would be useful to be able to override the label in default.yml. For example, you have a staging shc, and want to differentiate via the DMC. For more accurate testing, want to share a default.yml across both staging and prod.


jmervine:
Thanks for getting this in!

PR #283, opened 2019-10-28
Cleanup splunkd.pid and mongod.lock runtime files before starting Splunk

mikedickey:
Recommended by @jmervine


jmervine:
This is great. We take a slightly less elegant brute force approach, running the following tasks as the first thing done as part of our pre-install hook
```yaml
    - name: CUSTOM Remove old splunk pid
      command: rm -rf /opt/splunk/var/run/splunk/splunk.pid
      become: yes
    - name: CUSTOM Remove old kvstore lock file
      command: rm -rf /opt/splunk/var/lib/splunk/kvstore/mongo/mongod.lock
      become: yes
```

This should probably do the trick though. 👍

PR #288, opened 2019-10-30
Tod package install tests

tod-uma:
I've added molecule tests to check the package install options. I've used the docker images that I know work. If you have a suggestion of a different image, I'm fine with that. Also, found one bug in the .deb installer that I fixed.


bb03:
@tod-uma you are my new favorite person

PR #296, opened 2019-11-11
Fixed a bug causing splunk.apps_location to be ignored

mikedickey:
Fixed a bug where the value of splunk.apps_location provided via
defaults.yml was always being ignored by ansible playbooks

Fixed a bug when running docker container on Kubernetes, where
volume mount permissions were causing mongod to fail to startup
when persistent volume claims were reused with the following error:

permissions on /opt/splunk/var/lib/splunk/kvstore/mongo/splunk.key are too open


mikedickey:
I believe this fixes https://github.com/splunk/docker-splunk/issues/276

mikedickey:
Based on further testing, there still appears to be an issue with a circular dependency between initialization of SHC and startup of the deployer. So when apps_location is provided, both of these roles hang at startup. Still working on a fix for that...

PR #297, opened 2019-11-12
Fix #291  Enable HEC via master-apps in indexer cluster

### LINKED ISSUE #291: In a clustered ENV hec token should not be placed directly on indexers
This allows for data loss where the input is enabled on a new indexer before indexes.conf has been deployed. When in indexer mode HEC tokens must be applied via master-apps

rfaircloth-splunk:


roles/splunk_common/tasks/set_as_hec_receiver.yml
```diff
@@ -1,46 +1,37 @@
...
+    section: "http://{{ splunk_hec_token }}"
+    option: disabled
+    value: "0"
+    mode: '0600'
+    backup: no
   ignore_errors: true
-  when: splunk_hec_input.status != 409
-  no_log: "{{ hide_password }}"
-
-- name: Enable SSL for HEC new-token
```
nwang92: We shouldn't remove the ability to configure SSL here

roles/splunk_common/tasks/set_as_hec_receiver.yml
```diff
@@ -1,46 +1,37 @@
...
-    user: "{{ splunk.admin_user }}"
-    password: "{{ splunk.password }}"
-    validate_certs: false
-    body:
-      name: "splunk_hec_token"
-      token: "{{ splunk.hec_token }}"
-    body_format: "form-urlencoded"
-    status_code: 201,409
-    timeout: 10
+  ini_file:
```
nwang92: If you move to `ini_file`, we either need to do this step before Splunk starts, or you need to add a `notify` trigger here as to restart Splunk after the provisioning has completed. I'd prefer the former because unnecessary restarts add delay to the time-to-value from the moment users start their containers to when they can actually begin using Splunk.

roles/splunk_cluster_master/tasks/main.yml
```diff
@@ -11,6 +11,14 @@
   when:
     - splunk.apps_location
 
+- include_tasks: ../../splunk_common/tasks/set_as_hec_receiver.yml
+  vars:
+    splunk_hec_inputs_app_path: "apps/splunk_httpinput/local"
```
nwang92: We should already have a `splunk.app_paths.httpinput` path for this (minus the `local`)

roles/splunk_common/tasks/set_as_hec_receiver.yml
rfaircloth-splunk: your right I will add it back

bb03:
Hi @rfaircloth-splunk, are you having trouble making progress?  We would be happy to help answer any questions so you can submit this PR.

nwang92:
Going to close this PR because it's a little out-of-date and there's a discussion https://github.com/splunk/splunk-ansible/issues/439 on how we should be handling this going forward.

The HEC setup has been modified a lot since this PR was first issued. In a distributed topology, we still don't push it from cluster-master -> peers, but the `splunk.hec.*` settings are very tailored to simply 1 token and getting data in quickly. For larger, production deployments, you're likely to have >1 token which is (currently) beyond the scope of this parameter. If those tokens/HEC settings are bundled as an app, then you get all the goodness of master-apps and bundle pushing.

PR #299, opened 2019-11-13
fix issue of app bundle from index cluster master to peers.

hzhzhang:
Reorder the steps: prepare the apps and put them under folder "master-apps",  then mark the cluster master node and init the cluster.


nwang92:
If I understand these changes, it seems like the cluster-master initialization will now handle the bundle push to all of the downstream peers?

What happens during updates to an existing indexer cluster? For instance, if I have an indexer cluster already formed with app `A,B,C` and I modify the cluster-master to include app `D`, what I expect to happen is that app `D` gets added to all the indexers. But in this case, if the indexer cluster is already initialized, we may not be able to rely on that mechanism to distribute the new app/s to the peers.

hzhzhang:
@nwang92 Thanks for your comment. Here are some clarifications:
**Reason of the code change**:  If we initialize the cluster muster first, then generate the apps to bundle under master-apps folder while the peers joining the cluster in parallel , sometimes the bundle will be broken. We find this issue when deploy Splunk_TA_ForIndexers of app-ess. The cluster mater shows the bundle finishes successfully but the app is not distributed to the peers. And it won't be synced to peers even when we push the bundle manually again.
**About the concern of newly added app**: This playbook is for the initially installed apps. Newly added apps in existing cluster-master, another bundle push action is still needed. But that's not what this playbook will take.  It is okay to keep the apply_cluster_bundle.yml but it won't be called anywhere now.
**Side effects of the code change**: The cluster master initialization step will hang to wait for the bundle apps being ready, which will postpone the index peers clustering step. But it won't take long.

nwang92:
@hzhzhang While this does fix whatever problems may exist on startup, we can't ignore how these ansible playbooks address aspects during the full lifecycle of each role, especially since there are users outside of Splunk who are using this. We should find a way to reintegrate the apply_cluster_bundle.yml if that's the only way apps can get updated in an indexer cluster.

hzhzhang:
@nwang92 In that case, we can keep the apply_cluster_bundle.yml file. Is that okay？

nwang92:
I think it still needs to be called, otherwise leaving the playbook file in effectively does nothing for end users. These playbooks should be idempotent and guarantee that all the apps exist in a deployment, regardless of first-time run or upgrade.

hzhzhang:
@hzhzhang  added calling apply_cluster_bundle.yml in main.yml Thanks for your explanation of the use case.

PR #312, opened 2019-11-29
Fix deployer hang when apps_location is defined

mikedickey:
The shc_ready module never returned true, which caused the deployer role to hang whenever apps_location was specified via default.yml.

It was expecting the list of search head peers to be one greater than the number of peers in the search heads group. Previously, the captain was not a member of this group and at some point we changed
it such that the captain was a member. This caused a regression in this script which went undetected until now.


mikedickey:
Fixes https://github.com/splunk/docker-splunk/issues/276

PR #314, opened 2019-12-09
Enable Forwarder monitoring

jmeixensperger:
Create a saved search for building forwarder assets (enables monitoring) and execute it. Users can also pass in a cron schedule for the search (defaults to 15 mins).

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -110,8 +110,8 @@ def getDefaultVars():
     defaultVars["splunk"]["role"] = os.environ.get('SPLUNK_ROLE', 'splunk_standalone')
     defaultVars["splunk_home_ownership_enforcement"] = False if os.environ.get('SPLUNK_HOME_OWNERSHIP_ENFORCEMENT', "").lower() == "false" else True
     defaultVars["hide_password"] = True if os.environ.get('HIDE_PASSWORD', "").lower() == "true" else False
-    defaultVars["enable_forwarder_monitoring"] = os.environ.get('ENABLE_FORWARDER_MONITORING', False)
-    defaultVars["build_asset_schedule"] = os.environ.get('BUILD_ASSET_SCHEDULE', '18 * * * *')
+    defaultVars["dmc_forwarder_monitoring"] = os.environ.get('DMC_FORWARDER_MONITORING', True)
+    defaultVars["dmc_asset_interval"] = os.environ.get('DMC_ASSET_INTERVAL', '18 * * * *')
     #If the value is set, we would use that, otherwise, return True
     defaultVars["splunk"]["preferred_captaincy"] = False if os.environ.get('SPLUNK_PREFERRED_CAPTAINCY', "").lower() == "false" else True
     defaultVars["splunk"]["hostname"] = os.environ.get('SPLUNK_HOSTNAME', socket.getfqdn())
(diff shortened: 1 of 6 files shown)


inventory/environ.py
```diff
@@ -110,6 +110,8 @@ def getDefaultVars():
     defaultVars["splunk"]["role"] = os.environ.get('SPLUNK_ROLE', 'splunk_standalone')
     defaultVars["splunk_home_ownership_enforcement"] = False if os.environ.get('SPLUNK_HOME_OWNERSHIP_ENFORCEMENT', "").lower() == "false" else True
     defaultVars["hide_password"] = True if os.environ.get('HIDE_PASSWORD', "").lower() == "true" else False
+    defaultVars["dmc_forwarder_monitoring"] = os.environ.get('DMC_FORWARDER_MONITORING', False)
+    defaultVars["dmc_asset_interval"] = os.environ.get('DMC_ASSET_INTERVAL', '3,18,33,48 * * * *')
```
bb03: Make sure to add these two variables to the docs and document the cron schedule for the DMC assets.  15 minutes seems like a pretty good interval, but I can definitely see cases where customers would not want to run it this frequently.

jmeixensperger: 15 minutes was the default option through the UI, so I wanted to keep it consistent.

roles/splunk_common/tasks/enable_forwarder_monitoring.yml
```diff
@@ -0,0 +1,27 @@
+- name: Enable forwarder monitoring
+  ini_file:
+    path: "{{ splunk.home }}/etc/apps/splunk_monitoring_console/local/savedsearches.conf"
+    section: "DMC Forwarder - Build Asset Table"
+    option: "{{ item.key }}"
+    value: "{{ item.value }}"
+  with_dict: {"cron_schedule": "{{ dmc_asset_interval }}", "alert.track": 0, "disabled": 0,
```
bb03: Looking at this again, I don't think we need to update the visualizations section here, so we can probably take it out.  Shutting off the run_on_startup we should probably also set through the REST endpoint, and then we can likely avoid triggering the restart handler.

jmeixensperger: I looked back at the default savedsearches.conf file and you're correct, the visualizations and run_on_startup settings are already set to these values.
The one thing I'm unsure about is the restart handler because splunk was having trouble finding the search unless I triggered a restart. Keeping run_on_startup disabled also keeps the behavior consistent with this process through the UI.

bb03: Okay, sounds good.  Lets undo the visualization update but we can keep the restart trigger

jmeixensperger: I wasn't able to get it working without the restart trigger, but I removed the settings from the request body that are already assigned correctly.

roles/splunk_common/tasks/enable_forwarder_monitoring.yml
```diff
@@ -0,0 +1,22 @@
+- name: Enable forwarder monitoring
+  ini_file:
+    path: "{{ splunk.home }}/etc/apps/splunk_monitoring_console/local/savedsearches.conf"
+    section: "DMC Forwarder - Build Asset Table"
+    option: "{{ item.key }}"
+    value: "{{ item.value }}"
+  with_dict: {"cron_schedule": "{{ dmc_asset_interval }}", "alert.track": 0, "disabled": 0}
+  notify:
```
nwang92: Is this something we can also make a REST call to avoid the restart?

jmeixensperger: I tried out the REST call but was unable to even populate the local/savedsearches.conf, even with a restart.

PR #318, opened 2019-12-11
autodetect service_name for forwarder and allow override. systemd tests.

tod-uma:
I started the branch because I waned the splunkforwarder to be able to use the cli boot-start command on systemd systems. However I ran into the fact that by default the forwarder uses SplunkForwarder.service and enterprise uses Splunkd. Along the way I tried to make it so someone can manually specify the service name and tried to centralize the detection of systemd, etc. I also added molecule tests for systemd.

### DIFF
--- a/roles/splunk_common/handlers/restart_splunk.yml
+++ b/roles/splunk_common/handlers/restart_splunk.yml
@@ -9,9 +9,9 @@
   delay: "{{ delay_num }}"
   when: not splunk.enable_service
 
-- name: "Restart the splunkd service - Via systemd"
+- name: "Restart the splunkd service - Via Linux systemd or init"
   service:
-    name: "{% if pid1.stdout.find('systemd') != -1 %}Splunkd{% else %}{{ splunk.service_name }}{% endif %}"
+    name: "{{ splunk_service_name }}"
     state: restarted
   when: splunk.enable_service and ansible_system is match("Linux")
   become: yes
@@ -19,7 +19,7 @@
 
 - name: "Restart the splunkd service - Via windows system"
   win_service:
-    name: splunkd
+    name: splunk_service_name
     state: restarted
   when: splunk.enable_service and not ansible_system is match("Linux")
 
--- a/roles/splunk_common/tasks/enable_service.yml
+++ b/roles/splunk_common/tasks/enable_service.yml
@@ -1,8 +1,4 @@
 ---
-- name: "Retrieve PID 1 process information (Linux)"
-  command: "ps 1"
-  register: pid1
-  when:  ansible_system is match("Linux")
 
 - name: "Retrieve Splunk version"
   command: "{{ splunk.exec }} version --accept-license --answer-yes --no-prompt"
@@ -22,7 +18,7 @@
   command: "{{ splunk.exec }} enable boot-start -systemd-managed 1 -user {{ splunk.user }} --accept-license --answer-yes --no-prompt"
   when:
     - ansible_system is match("Linux")
-    - pid1.stdout.find("systemd") != -1
+    - splunk_systemd
     - installed_splunk_version[0] is version("7.2.2", ">=")
 
 # Using service file approach for systemd rather than 'boot-start' with
@@ -31,7 +27,7 @@
 - name: "Copy Splunkd unit file - Linux (systemd)"
   template:
     src: Splunkd.service.j2
-    dest: /etc/systemd/system/Splunkd.service
+    dest: "/etc/systemd/system/{{ splunk_service_name }}"
     owner: "{{ privileged_user }}"
     group: "{{ privileged_user }}"
     mode: 0644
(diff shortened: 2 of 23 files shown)


roles/splunk_common/tasks/enable_service.yml
```diff
@@ -31,33 +27,39 @@
...
     mode: 0644
   become: yes
   become_user: "{{ privileged_user }}"
   when:
     - ansible_system is match("Linux")
-    - pid1.stdout.find("systemd") != -1
+    - splunk_systemd
     - installed_splunk_version[0] is version("7.2.2", "<")
 
+- name: "Disable cgroup systemd settings in Docker"
```
nwang92: Hmm I was going to suggest this does not get included in here, but it seems like 8.0.0 has decided to deviate from the systemd unit file from 7.3.x: https://docs.splunk.com/Documentation/Splunk/8.0.0/Admin/RunSplunkassystemdservice

The docs seem to suggest that upon startup, Splunk 8.0.0 will automatically correct itself - so I'm still not sure if we should be responsible for doing it here, but it's something worth testing. There might also be some impact to workload management here as well? @bb03 
 
Also cc'ing @rcabral-splunk since this will directly impact your work as well.

tod-uma: oh yeah sorry. I intended to put a comment in there that I couldn't find a way to get that to work at least in my setup. I'm still kinda new to the whole docker thing so it could just be I don't have something set up right.

roles/splunk_common/tasks/get_facts.yml
```diff
@@ -123,3 +123,26 @@
...
+  set_fact:
+    splunk_systemd: "{{ pid1.stdout.find('systemd') != -1 }}"
+  when:
+    - ansible_system is match("Linux")
+
+# usually we can auto detect what the service name should be, but sometimes
+# we want to manually specify it. This is where we detect it was manually set.
+- name: "Setting service_name fact from config"
+  set_fact:
+    splunk_service_name: "{{ splunk.service_name }}"
```
rcabral-splunk: My tests in a non-dockerized/systemd Splunk installation failed here because I don't specify `service_name` parameter.
This fact will always be set to `splunk` by default if it is not manually overridden by the user. See service_name in [splunk_defaults_linux.yml](../../blob/develop/inventory/splunk_defaults_linux.yml#L76). In order to support this parameter override, maybe we should consider removing it from [splunk_defaults_linux.yml](../../blob/develop/inventory/splunk_defaults_linux.yml).
@nwang92 , @bb03?

roles/splunk_common/tasks/enable_service.yml
rcabral-splunk: I'll test that in my scenario with different enterprise versions because I'm not running Splunk as a container.

bb03: @rcabral-splunk How did the test go?

rcabral-splunk: I've tested 7.3.3 and 8.0.0 non-containerized versions. This particular change did not impact on anything, although the comments at https://github.com/splunk/splunk-ansible/pull/318#discussion_r357479197 are still a blocker

roles/splunk_common/tasks/get_facts.yml
rcabral-splunk: service_name also needs to be remove from `environ.py` to make this to work.

tod-uma: It shouldn't have failed there. It should have moved on to the next task which should automatically set the service name based on if the system is running systemd and/or using the universal forwarder. Can you send me the error?

roles/splunk_common/handlers/restart_splunk.yml
```diff
@@ -9,17 +9,17 @@
...
+    name: "{{ splunk_service_name }}"
     state: restarted
   when: splunk.enable_service and ansible_system is match("Linux")
   become: yes
   become_user: "{{ privileged_user }}"
 
 - name: "Restart the splunkd service - Via windows system"
   win_service:
-    name: splunkd
+    name: splunk_service_name
```
nwang92: This should be `"{{ splunk_service_name }}"`?

roles/splunk_common/tasks/get_facts.yml
```diff
@@ -129,3 +130,26 @@
     # standalone instances UNLESS they are specified by forward-servers
     - "'splunk_search_head' not in groups"
 
+- name: "Retrieve PID 1 process information (Linux)"
```
nwang92: Might be a duplicate of lines 8-12?

roles/splunk_common/tasks/install_splunk_yum.yml
```diff
@@ -1,7 +1,12 @@
+# added the environment section due to bug:
+# https://github.com/ansible/ansible/issues/18979
```
nwang92: cc @hendolim - wondering if this might also apply to https://github.com/splunk/docker-splunk/issues/303

roles/splunk_common/tasks/main.yml
```diff
@@ -32,6 +32,12 @@
     - splunk.allow_upgrade is defined
     - splunk.allow_upgrade | bool
 
+- name: "Set splunk-launch.conf options"
```
nwang92: Ah I didn't see this in this PR initially - this feature should be added already in line 77:
```
- include_tasks: set_launch_conf.yml
  when:
    - "'launch' in splunk and splunk.launch"
```

roles/splunk_common/tasks/get_facts.yml
nwang92: cc @rcabral-splunk

roles/splunk_common/tasks/main.yml
tod-uma: Yeah I'm learning to make my PR's smaller. I think I added this while someone else was working on the same thing and since my method was rolled into this big PR it wasn't noticed.

nwang92: No worries, appreciate the contributions :)

roles/splunk_common/tasks/get_facts_service_name.yml
```diff
@@ -0,0 +1,32 @@
...
+
+- name: "Setting service_name to splunkd"
+  set_fact:
+    splunk_service_name: "splunkd"
+  when:
+    - splunk_service_name is not defined
+    - splunk.enable_service and ansible_system is match("Linux")
+    - splunk_systemd is False
+
+- name: "Setting service_name to splunkd"
```
nwang92: It seems like this and the above task could be combined?

roles/splunk_common/tasks/get_facts_service_name.yml
```diff
@@ -0,0 +1,32 @@
+---
```
nwang92: Kinda leaning towards handling Linux vs. Windows // Splunk vs UF service naming in the `inventory/splunk*_defaults_*.yml` files since that matrix of statically-defined per-platform, per-product changes already exist there. 

Also it may be nicer to do it there so you don't spend Ansible-provisioning-time setting up variables. I might take a stab at that later though.

PR #322, opened 2019-12-13
do not add search peers in here if create multisite topology

kyanite:


roles/splunk_search_head/tasks/main.yml
```diff
@@ -9,11 +9,13 @@
 - include_tasks: peer_cluster_master.yml
   when:
     - splunk.indexer_cluster
+    - splunk.site is not defined
```
bb03: I would change this to check for "splunk.multisite_master", because people can define a single site with the prefix and not declare a multisite topology.  There are some configurations where people define a site so they can test certain configurations, but they don't set it as multisite.  Does this make sense?

kyanite: Make sense if our customers need such use case

bb03: Can you make that change to multisite_master and verify please?  Then I can merge it.

kyanite: but looks like multisite_master just be passed when execute multisite.yml, so it means when we execute site.yml, splunk.multisite_master is always not defined

PR #328, opened 2019-12-13
Using REST for 6.x.x and later packages

bb03:


arctan5x:
Code is very well written for readability. Kudos!

PR #331, opened 2019-12-18
Fixes for SHC expansion & resiliency

mikedickey:
Moved splunk_search_head_captain fact from the top-level site.yml to the main.yml of the splunk_search_head role. It was out of place in site.yml and is only used by plays within that role.

Added "Add new member to SHC" which can be used to append new search heads to an existing cluster. Previously, it was not possible to add new members.

Sanity check to allow destructive sync to fail if member is the captain. This was previously causing the playbooks to repeatedly retry the command until failing, whenever the captaincy was transferred to a member other than the original one, and then re-run on that member (i.e. due to restart).

Reduced destructive sync retry delay from 60 seconds (which seemed quite excessive) to 10 seconds.

### DIFF
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -84,5 +84,5 @@
   failed_when: task_result.rc !=0 and "this instance is the captain" not in task_result.stderr
   until: task_result.rc == 0 or "this instance is the captain" in task_result.stderr
   retries: "{{ retry_num }}"
-  delay: 10
+  delay: 60
   no_log: "{{ hide_password }}"
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -70,14 +70,28 @@
     - Restart the splunkd service
   no_log: "{{ hide_password }}"
 
+- name: Add new member to SHC
+  command: "{{ splunk.exec }} add shcluster-member -current_member_uri {{ cert_prefix }}://{{ groups['splunk_search_head_captain'][0] }}:{{ splunk.svc_port }} -auth {{ splunk.admin_user }}:{{ splunk.password }}"
+  become: yes
+  become_user: "{{ splunk.user }}"
+  when: not splunk_search_head_captain | bool
+  register: task_result
+  changed_when: task_result.rc == 0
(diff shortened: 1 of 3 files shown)


roles/splunk_search_head/tasks/search_head_clustering.yml
```diff
@@ -70,14 +70,28 @@
     - Restart the splunkd service
   no_log: "{{ hide_password }}"
 
+- name: Add new member to SHC
+  command: "{{ splunk.exec }} add shcluster-member -current_member_uri {{ cert_prefix }}://{{ groups['splunk_search_head_captain'][0] }}:{{ splunk.svc_port }} -auth {{ splunk.admin_user }}:{{ splunk.password }}"
```
bb03: My only question here is whether or not ```groups['splunk_search_head_captain'][0]``` will always be defined when we execute this block.  It should, I think.

What kinds of configurations did you test these changes with?

mikedickey: As I understand it, this would be populated from `SPLUNK_SEARCH_HEAD_CAPTAIN_URL` environment variable. According to [current docs](https://github.com/splunk/splunk-ansible/blob/develop/docs/ADVANCED.md), this is required to enable SHC.

I tested by spinning up a new cluster, killing off and restarting various nodes (including captain), expanding cluster, killing off and restarting expanded nodes.

bb03:
@mikedickey The reason those numbers were so large were from practical concerns regarding how long it may take for a cluster master to initialize, especially in large scale environments (I think 1K indexers and 300 search heads).  We should probably make this value adjustable from the end user.

nwang92:
It appears the increased delay was for splunk-ansible on Windows: https://github.com/splunk/splunk-ansible/pull/267/files

Maybe we can expand the delay but not as generously 10s -> 15 or 20s as opposed to 60s?

mikedickey:
I reverted my delay change for "Destructive sync search head" so that we can consider and merge this independent of that change. I'm going to work on another PR to take a stab at all these delay/retry issues..

bb03:
So one of the things that @nwang92 and I were talking about is why we have that delay here in the first place and why are the linux systems affected by a value that should apply to windows.  Considering that 90% or more of our internal usage is for linux only, theres no good reason why linux should suffer through the same delays.

mikedickey:
@bb03 I agree and have some other issues as well with our retry/delay settings. I have a second PR ready to try to address these, but it's built on top of the bug fixes in this one. So I'll push that as soon as this one merges.

PR #332, opened 2019-12-19
Global updates to retries and delay

mikedickey:
This PR attempts to make task retry behaviour more consistent. Many tasks had specific hard-coded values for "retries" and "delay" which contradicted the global configuration parameters for these settings. I suspect that these were written before configurable parameters were introduced, and over time copied to new tasks which grew inconsistency.

This PR removes any specific values for these parameters, ensuring that the configuration parameters are universally respected.

This also renamed the global parameters "delay_num" to "retry_delay" which is more consistent with "retry_num" and not misleading. This parameter is used to specify the delay in seconds between retries, so "_num" seemed inappropriate as a name.

A previous PR (https://github.com/splunk/splunk-ansible/pull/267/files) introduced alternative delay values for the tasks that perform destructive sync and wait for the instance to be available. This was reported as an issue specific for large Windows deployments, but the values were enlarged
for all platforms.

To better address this exception more explicitly, I added to new configuration parameters: wait_for_splunk_retry_num and shc_sync_retry_num. I believe it is far better to have larger retry numbers versus larger delays, since the latter has a broadly negative impact across the board, even when things are efficient. Enlarging the retry number yields a similar result, without the negative side effects.

Continuing on preference for retries versus delay, I replaced the shc_bootstrap_delay configuration parameter with shc_sync_retry_num.

Also to address general apparent issues with slowness on Windows, I made the default retry_delay larger (10 seconds) versus Linux (3 seconds). This probably isn't necessary, and I do fear we are just masking out other underlying problems by having these larger delays, and perhaps that is something that should be investigated more as part of a separate effort. In any case, the new values for destructive sync and wait for instance tasks should provide a consistent buffer as with previous behavior.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -34,7 +34,7 @@ While developing new playbooks that require remote Splunk-to-Splunk connectivity
   register: set_indexer_as_peer
   until: set_indexer_as_peer.rc == 0 or set_indexer_as_peer.rc == 24
   retries: "{{ retry_num }}"
-  delay: 3
+  delay: "{{ retry_delay }}"
   changed_when: set_indexer_as_peer.rc == 0
   failed_when: set_indexer_as_peer.rc != 0 and 'already exists' not in set_indexer_as_peer.stderr
   notify:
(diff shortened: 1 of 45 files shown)


mikedickey:
This is a follow-up to https://github.com/splunk/splunk-ansible/pull/331

nwang92:
LGTM - this likely breaks backwards compatibility with any `default.yml` generated from a previous versions of the docker image. I'm ok with this because there are other changes in develop (but not in master/released to the wild yet) that also require regeneration. But adding a comment here so whoever does the next release makes sure to include this in release notes.

Some food for thought, but maybe we should take advantage of Github labels to mark changes like this?

PR #344, opened 2020-01-07
Moving rectify search/replication factors into python

nwang92:
Moving all rectification of search/replication factors into environ.py.

This should solve immediate issues with M2 deployment type, although I think there may be inherent flaws with defining search/replication factors here. I'm still working my way through the docs, but in the case of replication factor, it seems like there are dependencies on what the replication factor is defined to be for each individual site:
> If there are some non-explicit sites, then the total value must be at least the sum of all explicit site and origin values.

This makes it very easy to incorrectly set the `site_search_factor_total` because the `default.yml` supplied is not aware of any cluster's site topology other than it's own. Currently we can get by with just min/max, but this is not a very robust solution and it makes not only error correction difficult but it also makes any fail-fast/error-raising conditions harder to infer.

I may take another stab at this later, but putting a PR out for some feedback.

Docs:
https://docs.splunk.com/Documentation/Splunk/8.0.1/Indexer/Sitesearchfactor
https://docs.splunk.com/Documentation/Splunk/8.0.1/Indexer/Sitereplicationfactor

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -148,8 +148,16 @@ def getIndexerClustering(vars_scope):
     idxc_vars["secret"] = os.environ.get("SPLUNK_IDXC_SECRET", idxc_vars.get("secret"))
     # Rectify cluster replication + search factors
     indexer_count = len(inventory.get("splunk_indexer", {}).get("hosts", []))
-    idxc_vars["replication_factor"] = min(indexer_count, int(idxc_vars.get("replication_factor", 0)))
-    idxc_vars["search_factor"] = min(idxc_vars["replication_factor"], int(idxc_vars.get("search_factor", 0)))
+    # Replication factor should be <= number of indexers
+    replf = int(idxc_vars.get("replication_factor", 0))
+    if replf > indexer_count:
+        raise Exception("Cannot meet replication factor of {}, not enough peers available".format(replf))
+    idxc_vars["replication_factor"] = replf
+    # Search factor should be <= replication factor
+    searchf = int(idxc_vars.get("search_factor", 0))
+    if searchf > replf:
+        raise Exception("Cannot meet search factor of {}, must be at most {} (replication factor)".format(searchf, replf))
+    idxc_vars["search_factor"] = searchf
 
 def getSearchHeadClustering(vars_scope):
     """
@@ -162,7 +170,11 @@ def getSearchHeadClustering(vars_scope):
     shc_vars["secret"] = os.environ.get("SPLUNK_SHC_SECRET", shc_vars.get("secret"))
     # Rectify SHC replication factor
     sh_count = len(inventory.get("splunk_search_head", {}).get("hosts", []))
-    shc_vars["replication_factor"] = min(sh_count, int(shc_vars.get("replication_factor", 0)))
+    # Replication factor should be <= number of search heads
+    replf = int(shc_vars.get("replication_factor", 0))
+    if replf > sh_count:
+        raise Exception("Cannot meet replication factor of {}, not enough search heads available".format(replf))
+    shc_vars["replication_factor"] = replf
 
 def getMultisite(vars_scope):
     """
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -95,6 +95,8 @@ def getDefaultVars():
 
     getIndexerClustering(defaultVars)
     getSearchHeadClustering(defaultVars)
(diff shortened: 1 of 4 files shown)


inventory/environ.py
```diff
@@ -116,12 +118,6 @@ def getDefaultVars():
     if os.environ.get("SPLUNK_PREFERRED_CAPTAINCY", "").lower() == "false":
         defaultVars["splunk"]["preferred_captaincy"] = False
     defaultVars["splunk"]["hostname"] = os.environ.get('SPLUNK_HOSTNAME', socket.getfqdn())
-    #When sites are specified, assume multisite
```
nwang92: I removed this section because I wasn't sure what it was used for.

inventory/environ.py
```diff
@@ -140,6 +136,10 @@ def getIndexerClustering(vars_scope):
     idxc_vars = vars_scope["splunk"]["idxc"]
     idxc_vars["label"] = os.environ.get("SPLUNK_IDXC_LABEL", idxc_vars.get("label"))
     idxc_vars["secret"] = os.environ.get("SPLUNK_IDXC_SECRET", idxc_vars.get("secret"))
+    # Rectify cluster replication + search factors
+    indexer_count = len(inventory.get("splunk_indexer", {}).get("hosts", []))
+    idxc_vars["replication_factor"] = min(indexer_count, int(idxc_vars.get("replication_factor", 0)))
```
nwang92: This may be a good spot to raise an exception to let users know the user-defined replication factor cannot be met instead of just correcting it for them?

inventory/environ.py
```diff
@@ -150,13 +150,16 @@ def getSearchHeadClustering(vars_scope):
     shc_vars = vars_scope["splunk"]["shc"]
     shc_vars["label"] = os.environ.get("SPLUNK_SHC_LABEL", shc_vars.get("label"))
     shc_vars["secret"] = os.environ.get("SPLUNK_SHC_SECRET", shc_vars.get("secret"))
+    # Rectify SHC replication factor
+    sh_count = len(inventory.get("splunk_search_head", {}).get("hosts", []))
+    shc_vars["replication_factor"] = min(sh_count, int(shc_vars.get("replication_factor", 0)))
```
nwang92: This may be a good spot to raise an exception to let users know the user-defined replication factor cannot be met instead of just correcting it for them?

inventory/environ.py
```diff
@@ -166,18 +169,20 @@ def getMultisite(vars_scope):
...
+    # TODO: Split this into its own splunk.multisite.* section
+    splunk_vars["multisite_master_port"] = int(os.environ.get("SPLUNK_MULTISITE_MASTER_PORT", splunk_vars.get("multisite_master_port", 8089)))
+    splunk_vars["multisite_replication_factor_origin"] = int(os.environ.get("SPLUNK_MULTISITE_REPLICATION_FACTOR_ORIGIN", splunk_vars.get("multisite_replication_factor_origin", 1)))
+    splunk_vars["multisite_replication_factor_total"] = int(os.environ.get("SPLUNK_MULTISITE_REPLICATION_FACTOR_TOTAL", splunk_vars.get("multisite_replication_factor_total", 1)))
+    splunk_vars["multisite_replication_factor_total"] = max(splunk_vars["multisite_replication_factor_total"], splunk_vars["idxc"]["replication_factor"])
+    splunk_vars["multisite_search_factor_origin"] = int(os.environ.get("SPLUNK_MULTISITE_SEARCH_FACTOR_ORIGIN", splunk_vars.get("multisite_search_factor_origin", 1)))
+    splunk_vars["multisite_search_factor_total"] = int(os.environ.get("SPLUNK_MULTISITE_SEARCH_FACTOR_TOTAL", splunk_vars.get("multisite_search_factor_total", 1)))
+    splunk_vars["multisite_search_factor_total"] = max(splunk_vars["multisite_search_factor_total"], splunk_vars["idxc"]["search_factor"])
```
nwang92: Using max here because the error exhibited was:
```
Invalid site_search_factor, Reason: total should be greater than or equal to search_factor
```

inventory/environ.py
bb03: @nwang92 Yeah, I think this is the right spot for it

nwang92:
Per discussion with @bb03, there's a good chance this may be too disruptive at the moment because the only way to configure search/replication factor is with the `default.yml`. I'm going to change this PR to expose new envvars to support this, and move exception raising on invalid configurations to a future PR.

PR #348, opened 2020-01-10
More refactoring and unit-tests

nwang92:
There seemed to be lots of layers of these merge_dict() calls, but I've reduced it down in the following order:
1. base defaults.yml shipped in this repo (`inventory/splunk_defaults_*.yml`)
2. "baked" defaults.yml (from config.baked - reading file on disk)
3. "env" defaults.yml (from config.env.var - either file on disk or fetched from a URL)
4. "host" defaults.yml (from config.host.url - fetched from a URL)
5. os.environ vars

Previously, the last defaults file ("host" defaults.yml) was called from `loadHostVars()` which actually ran after the os.environ overrides. I didn't think that was a great pattern to continue because it seems unnecessarily complicated to parse YAMLs from files/URLs, override with env vars, then override once more with some special YAML that comes from some URL.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -34,6 +34,7 @@ urllib3.disable_warnings()
 HERE = os.path.dirname(os.path.normpath(__file__))
 _PLATFORM = platform.platform().lower()
 PLATFORM = "windows" if ("windows" in _PLATFORM or "cygwin" in _PLATFORM) else "linux"
+HOSTNAME = os.uname()[1]
 JAVA_VERSION_WHITELIST = frozenset(("oracle:8", "openjdk:8", "openjdk:9", "openjdk:11"))
 
 roleNames = [
@@ -91,18 +92,18 @@ def getSplunkInventory(inventory, reName=r"(.*)_URL"):
     if os.path.isfile("/.dockerenv"):
         if "localhost" not in inventory["all"]["children"]:
             inventory["all"]["hosts"].append("localhost")
-        inventory["_meta"]["hostvars"]["localhost"] = loadHostVars(inventory["all"]["vars"],
-                                                                   os.uname()[1], PLATFORM)
+        inventory["_meta"]["hostvars"]["localhost"] = inventory["all"]["vars"]
         inventory["_meta"]["hostvars"]["localhost"]["ansible_connection"] = "local"
 
 def getDefaultVars():
     """
-    Load all splunk-ansible defaults and perform overwrites based on 
+    Load all splunk-ansible defaults and perform overwrites based on
     environment variables to return a consolidated inventory object
     """
-    defaultVars = loadDefaultSplunkVariables()
+    defaultVars = loadDefaults()
     overrideEnvironmentVars(defaultVars)
 
+    getSplunkPaths(defaultVars)
     getIndexerClustering(defaultVars)
     getSearchHeadClustering(defaultVars)
     getMultisite(defaultVars)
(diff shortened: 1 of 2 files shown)


inventory/environ.py
```diff
@@ -294,10 +308,6 @@ def overrideEnvironmentVars(vars_scope):
     vars_scope["ansible_pre_tasks"] = os.environ.get("SPLUNK_ANSIBLE_PRE_TASKS", vars_scope["ansible_pre_tasks"])
     vars_scope["ansible_post_tasks"] = os.environ.get("SPLUNK_ANSIBLE_POST_TASKS", vars_scope["ansible_post_tasks"])
     vars_scope["cert_prefix"] = os.environ.get("SPLUNK_CERT_PREFIX", vars_scope.get("cert_prefix", "https"))
-    vars_scope["splunk"]["opt"] = os.environ.get('SPLUNK_OPT', vars_scope["splunk"]["opt"])
-    vars_scope["splunk"]["home"] = os.environ.get('SPLUNK_HOME', vars_scope["splunk"]["home"])
-    vars_scope["splunk"]["exec"] = os.environ.get('SPLUNK_EXEC', vars_scope["splunk"]["exec"])
-    vars_scope["splunk"]["pid"] = os.environ.get('SPLUNK_PID', vars_scope["splunk"]["pid"])
```
nwang92: This was moved into `getSplunkPaths()` but it seems a bit dangerous to change these independently (at the end of the day, splunk.exec depends on splunk.home). But I'm guessing most people never touch this so maybe not very important to fix now.

inventory/environ.py
```diff
@@ -307,7 +317,6 @@ def overrideEnvironmentVars(vars_scope):
     vars_scope["splunk"]["enable_service"] = os.environ.get('SPLUNK_ENABLE_SERVICE', vars_scope["splunk"]["enable_service"])
     vars_scope["splunk"]["service_name"] = os.environ.get('SPLUNK_SERVICE_NAME', vars_scope["splunk"]["service_name"])
     vars_scope["splunk"]["allow_upgrade"] = os.environ.get('SPLUNK_ALLOW_UPGRADE', vars_scope["splunk"]["allow_upgrade"])
-    vars_scope["splunk"]["build_location"] = os.environ.get('SPLUNK_INSTALLER', vars_scope["splunk"]["build_location"])
```
nwang92: I couldn't find any references to `SPLUNK_INSTALLER` in splunk-ansible or `docker-splunk`

inventory/environ.py
```diff
@@ -229,7 +229,9 @@ def getLicenses(vars_scope):
     if vars_scope["splunk"]["license_uri"] and '*' in vars_scope["splunk"]["license_uri"]:
         vars_scope["splunk"]["wildcard_license"] = True
     vars_scope["splunk"]["nfr_license"] = os.environ.get("SPLUNK_NFR_LICENSE", "/tmp/nfr_enterprise.lic")
-    vars_scope["splunk"]["ignore_license"] = bool(os.environ.get("SPLUNK_IGNORE_LICENSE"))
+    vars_scope["splunk"]["ignore_license"] = False
+    if os.environ.get("SPLUNK_IGNORE_LICENSE", "").lower() == "true":
```
bb03: I'm fine if we don't want to support other bool representations - i.e. 0,1 - but we should probably make sure that its documented that we only support "true" and "false"

nwang92: Sure thing, I'll make another PR for docs because I think there's a lot we need to publish for this next release. On this point, there's a couple other boolean-like checks like this we do throughout, might be useful to have a common function to address the truthy-like value during this parsing.

PR #350, opened 2020-01-13
check SET_SEARCH_PEERS before trying to add search peers

kyanite:


roles/splunk_search_head/tasks/main.yml
```diff
@@ -20,13 +20,13 @@
 - include_tasks: peer_cluster_master.yml
   when:
     - splunk.indexer_cluster
-    - splunk.multisite_master is not defined
+    - splunk.set_search_peers is defined
```
bb03: @kyanite If you're going to go this route, we need set_search_peers in addition to the multisite_master check - not replacing the multisite_master check

bb03: Also, please change the check so that it can preserve backwards compatibility

bb03: @kyanite Please make sure to update the tests as necessary for https://github.com/splunk/docker-splunk for this change to proceed. It will break the clustering test-scenarios

bb03: Okay, I believe I was wrong about the clustering scenarios.  Approving

inventory/environ.py
```diff
@@ -340,6 +340,8 @@ def overrideEnvironmentVars(vars_scope):
     vars_scope["splunk"]["enable_service"] = os.environ.get('SPLUNK_ENABLE_SERVICE', vars_scope["splunk"]["enable_service"])
     vars_scope["splunk"]["service_name"] = os.environ.get('SPLUNK_SERVICE_NAME', vars_scope["splunk"]["service_name"])
     vars_scope["splunk"]["allow_upgrade"] = os.environ.get('SPLUNK_ALLOW_UPGRADE', vars_scope["splunk"]["allow_upgrade"])
+    if "SET_SEARCH_PEERS" in os.environ:
+        vars_scope["splunk"]["set_search_peers"] = os.environ.get('SET_SEARCH_PEERS')
```
nwang92: I'd recommend adding the `SPLUNK_` prefix so we know this env var is used only for our purposes. 

Also, can we change this variable to a proper boolean? Using the existence/non-existence of a variable as a feature-flag has the potential to be more error-prone for future development.

kyanite: @nwang92 , you mean use SET_SPLUNK_SEARCH_PEERS?

nwang92: Yeah - something like `SPLUNK_SET_SEARCH_PEERS` works. I would change it so the default is always `True` because it's unlikely anyone will create a SHC but not peer the indexer tier. For instance:
```
vars_scope["splunk"]["set_search_peers"] = bool(vars_scope["splunk"].get("set_search_peers", True))
if os.environ.get("SPLUNK_SET_SEARCH_PEERS", "").lower() == "false":
    vars_scope["splunk"]["set_search_peers"] = False
```

nwang92: ^ I guess following the above, if it's a boolean we should include this as part of Ansible checks:
```
- splunk.set_search_peers is defined and splunk.set_search_peers | bool
```

kyanite: except multisite topology, SET_SPLUNK_SEARCH_PEERS always True

Issue #358, opened 2020-01-23
Support newer version of ESS


nwang92:
When running the latest version of ESS, there's a Python compatibility problem:
```
2020-01-22 00:24:19,090 p=94 u=ansible |  TASK [splunk_cluster_master : Execute bundle script] ***************************
2020-01-22 00:24:19,091 p=94 u=ansible |  fatal: [localhost]: FAILED! => {
    "changed": false,
    "cmd": [
        "/opt/splunk/bin/splunk",
        "cmd",
        "python",
        "/tmp/es_ta_for_indexers.py",
        "--password",
        "Chang3d!"
    ],
    "delta": "0:00:00.231805",
    "end": "2020-01-22 00:24:19.070679",
    "rc": 1,
    "start": "2020-01-22 00:24:18.838874"
}
STDERR:
Traceback (most recent call last):
  File "/tmp/es_ta_for_indexers.py", line 86, in <module>
    make_ta_for_indexers(args.username, args.password)
  File "/tmp/es_ta_for_indexers.py", line 50, in make_ta_for_indexers
    from app_maker.make_index_time_properties import makeIndexTimeProperties
  File "/opt/splunk/etc/apps/SA-Utils/bin/app_maker/make_index_time_properties.py", line 83
    temp_conf = f'\n[{conf}]\n'
                              ^
SyntaxError: invalid syntax
```

It seems like the default `./bin/splunk cmd python` is Python 2, and thus won't work with this script. We should check to make sure if Splunk version >= 8.0.x and ES version is >= 6.1.x that we use `./bin/splunk cmd python3` as the executor.

PR #360, opened 2020-01-23
Fixing ESS 6.1+ installation

nwang92:
Fix for https://github.com/splunk/splunk-ansible/issues/358


nwang92:
Decided to only check ESS version - if 6.1+ is used with an older Splunk version, that likely means it will be unsupported and I think that burden of responsibility will be on the customer.

Issue #349, opened 2020-01-11
Consider allowing configurable OPTIMISTIC_ABOUT_FILE_LOCKING=x


aegershman:
I'm running splunk heavy forwarder docker images as apps on Cloud Foundry to use them as HECs to forward logs to our corporate splunk receivers.

[I get the error, `Validating databases (splunkd validatedb) failed with code '1'.` Setting `OPTIMISTIC_ABOUT_FILE_LOCKING=1` in `SPLUNK_HOME/etc/splunk-launch.conf` allows the images to start and function](https://answers.splunk.com/answers/306998/why-am-i-getting-homepathoptsplunkvarlibsplunkaudi.html). (I don't need these HECs to actually write logs to disk or view the UI. I just need them to receiving/transforms/forwarding.)

In order to set this I have to fully overwrite `SPLUNK_HOME/etc/splunk-launch.conf` with just this line. Is there an alternative way to configure `OPTIMISTIC_ABOUT_FILE_LOCKING=x`, or could a configuration flag be enabled to allow setting this? thoughts welcome.

thanks for any time 👍

nwang92:
Hmm I wonder if this is partially related to how CloudFoundry runs containers? The only time I've hit this error is when `/opt/splunk/var` is not backed by an appropriate Docker volume. As in, you'd only get the error if you installed Splunk in a container like:
```
docker run -d -t ubuntu bash
```
but you wouldn't if you installed Splunk in a container like:
```
docker run -d -t -v /opt/splunk/var ubuntu bash
```

In the dockerfiles in docker-splunk, there's a spec in the Dockerfile to always bring up a local volume, even if you don't specify one explicitly: https://github.com/splunk/docker-splunk/blob/develop/splunk/common-files/Dockerfile#L74

I suppose maybe the storage-driver of the docker daemon would also have an impact, but I haven't done any tests to see if that's the case. At the moment, we only support `overlay2` per recommendation from folks at Docker. 

Anyways, not sure if that thinking-out-loud was any helpful :P Is there a way you can share your spec for how these images are deployed on CloudFoundry?

nwang92:
@bb03 and I had a discussion about this, thinking about it more it may make sense to open up ways to add content into `splunk-launch.conf`. In addition to `OPTIMISTIC_ABOUT_FILE_LOCKING`, there are other applications of modifying this file to support things like DFS, Java, etc.

[...]

PR #362, opened 2020-01-24
Adding support for setting variables in splunk-launch.conf

nwang92:
Should address https://github.com/splunk/splunk-ansible/issues/349 and other use cases where we want to set vars in splunk-launch.conf

Will add a test in docker-splunk once this gets merged.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -358,7 +358,7 @@ def getLaunchConf(vars_scope):
     # From environment variables
     settings = os.environ.get("SPLUNK_LAUNCH_CONF")
     if settings:
-        launch.update({k:v for k,v in [x.split("=") for x in settings.split(",")]})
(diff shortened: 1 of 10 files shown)


inventory/environ.py
```diff
@@ -329,10 +330,29 @@ def getSplunkApps(vars_scope):
...
+    """
+    Parse key/value pairs to set in splunk-launch.conf
+    """
+    launch = {}
+    if not "launch" in vars_scope["splunk"]:
+        vars_scope["splunk"]["launch"] = {}
+    # From default.yml
+    if type(vars_scope["splunk"]["launch"]) == dict:
+        launch.update(vars_scope["splunk"]["launch"])
+    # From environment variables
```
bb03: I want to definitely add documentation about how to use this.  Can you make a separate page
"Creating a custom splunk_launch.conf" or something like that - then fill out with the usage and an example?

nwang92: Can do

PR #365, opened 2020-01-27
Feature/asan builds

nwang92:
Adding a trigger to do some special provisioning when using ASan Splunk builds. This is mostly for internal usage, which is why I did not include a `splunk.asan` in any of the shipped default.ymls - although if enough people feel strongly about doing so I can make it happen.

I still need to do some testing on this, but theoretically this should allow developer debugging using something like:
```
docker run -e SPLUNK_ENABLE_ASAN=yes -e SPLUNK_BUILD_URL=http://host/splunk-unstripped.tgz -e SPLUNK_LAUNCH_CONF= LSAN_OPTIONS=exitcode=0:report_objects:1, ASAN_OPTIONS=detect_leaks=1:log_path=/opt/splunk/_asanoutput/splunk:halt_on_error=false:atexit=true:print_stats=true:detect_stack_use_after_return=true:debug=true ... splunk/splunk:edge
```


bb03:
@jrigassio-splunk lets hold off on documentation, because I don't see this being used externally.

Issue #359, opened 2020-01-23
Version check in enable_service.yml causes permission issue on $SPLUNK_HOME/etc/myinstall/splunkd.xml


scalatar:
**CentOS**: 
CentOS Linux release 7.7.1908 (Core)

**Python**: 
2.7.5

**Ansible**:
```
ansible 2.10.0.dev0
  config file = None
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /root/git/ansible/lib/ansible
  executable location = /root/git/ansible/bin/ansible
  python version = 2.7.5 (default, Aug  7 2019, 00:51:29) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```

**Assumptions**:

- Splunk installed as root
- Splunk intended to run as unprivileged user via systemd

The issue appears to be that a version check is done prior to `enable boot-start`. The result is that the file `splunkd.xml` is created under `$SPLUNK_HOME/etc/myinstall/splunkd.xml` and is not owned by the splunk user. As a result the Splunkd service does not start complaining about permissions to splunkd.xml.

**Steps to reproduce manually**:

1. Sudo to root `sudo su -`
2. Download splunk rpm to /tmp/splunk.rpm
3. Install splunk rpm `yum localinstall /tmp/splunk.rpm`
4. Check version of splunk `/opt/splunk/bin/splunk version --accept-license --answer-yes --no-prompt`
5. Verify that /opt/splunk/etc/myinstall/splunkd.xml exists and it not owned by splunk user `ls -l /opt/splunk/etc/myinstall/splunkd.xml`

A potential fix would be to ensure the Splunk version command is run as the splunk user. For example inside enable_service.yml the following code using `become_user` appears to resolve the problem:

`- name: "Retrieve Splunk version"
  become: yes
  become_user: "{{ splunk.user }}"
  command: "{{ splunk.exec }} version --accept-license --answer-yes --no-prompt"
  register: installed_splunk_version
  when: ansible_system is match("Linux")`

nwang92:
Thanks for catching this! I merged a fix into develop - I have some other ideas on how to avoid making this `./bin/splunk` call in the future, but I'll chalk that up to future optimizations.

PR #366, opened 2020-01-27
Ensuring version check is done by splunk.user

nwang92:
Looks like we missed a case where the `{{ splunk.exec }}` should be executed as the Splunk user.

https://github.com/splunk/splunk-ansible/issues/359


nwang92:
Now that I think about it, I think we have this variable in `get_facts.yml` from either the splunk manifest or the file name. I'll see if I can clean this up in another PR so that we don't even need to invoke `./bin/splunk version`

PR #370, opened 2020-01-30
Setting pass4symmkey for IDXC instead of SHC

alishamayor:


alishamayor:
@dexa187 Looks like it was setting the wrong pass4SymmKey, so I've updated it to use the indexer clustering one instead of the SHC one.

Issue #206, opened 2019-06-26
Assumption that passwords will be the same across instances when using indexer clustering


pgullette:
I'm working a distributed deployment, and in my case I was starting with the Splunk AMI for AWS. This AMI has splunk-ansible already installed which is great and saves some steps, but the behavior of using that AMI is to create the admin password for splunk based on the created instance id.

I modified the playbook to copy instance-specific default.yml files. Within each default.yml, I can define the unique parts such as password. I noticed when I did my deploy including an indexer cluster that it was failing during peering the search head to the cluster master. Specifically, in peer_cluster_master.yml, there is an include to wait for the cluster master instance. That task is using the in-scope splunk user and password, which in this case is the admin user and password for the search head. This obviously creates a 401: Unauthorized when the admin password on the CM is different than SH.

There are ways to solve this, but I was curious if this was an assumption about using this for a distributed setup.

nwang92:
Yeah at the moment we do assume common passwords across all instances of your Splunk Enterprise deployment. We're currently investigating ways to open up the ability to define usernames/passwords on a per-instance or per-role basis.

As far as the Splunk AMI for AWS goes, I imagine the possible workarounds would be:
* After instance launch, change the admin password from instance-id to something else before running ansible
* Configure user-defined password at initial launch? I'm not sure if this is possible, but I imagine this would be a nice feature (cc @rcabral-splunk)

It seems to me that if every Splunk EC2 gets a dynamic password (based on instance-id) without the ability to pre-define it, you'll likely never be able to use the Splunk AMI for AWS for any distributed environments. Not sure if that's by design, but assuming you are able to get a distributed environment brought up, it will be very brittle and error-prone when instances die and get recreated.

bb03:
I tried this with the docker-splunk plays.  Speciicially I took the play here: https://github.com/splunk/docker-splunk/blob/develop/test_scenarios/3idx3sh1cm.yaml and modified it so that each tier contained different passwords (i.e. ```SPLUNK_PASSWORD=${SPLUNK_PASSWORD_IDX}``` for the indexer environment variables, similar thing for the search head and cluster master).  A little surprisingly, the splunk environment I got back appears to be correct, even though I am seeing this error in my logs
```idx2    | TASK [splunk_indexer : Check Splunk instance is running] ***********************
idx2    | fatal: [localhost]: FAILED! => {
idx2    |     "attempts": 60, 
idx2    |     "cache_control": "private", 
[...]

PR #388, opened 2020-02-16
Don't require admin authentication when waiting for remote instance

mikedickey:
Updated wait_for_splunk_instance.yml so that the HTTP request it makes does not require authentication. When waiting for a remote instance, this created a dependency that the remote instance must have the same admin password as the local instance, which likely doesn't hold true outside of crafted test environments. Splunk itself has no such requirement, which is why it has and uses pass4SymmKey secrets.

Only the root endpoint appears to be unauthenticated in the REST API, so this prevents us from being able to query uptime of the remote instance and ensure it is > 10 seconds. I don't know why this check was put in place, but generally artificial wait times like this can be better handled via more specific checks and retry loops.

### DIFF
--- a/roles/splunk_common/tasks/wait_for_splunk_instance.yml
+++ b/roles/splunk_common/tasks/wait_for_splunk_instance.yml
@@ -1,15 +1,12 @@
 ---
 - name: Check Splunk instance is running
   uri:
-    url: "{{ cert_prefix }}://{{ splunk_instance_address }}:{{ splunk.svc_port }}/services/server/info?output_mode=json"
+    url: "{{ cert_prefix }}://{{ splunk_instance_address }}:{{ splunk.svc_port }}"
     method: GET
-    user: "{{ splunk.admin_user }}"
-    password: "{{ splunk.password }}"
     validate_certs: false
   register: task_response
   until:
     - task_response.status == 200
-    - lookup('pipe', 'date +"%s"')|int - task_response.json.entry[0].content.startup_time > 10
   retries: "{{ wait_for_splunk_retry_num }}"
   delay: "{{ retry_delay }}"
   ignore_errors: true


roles/splunk_common/tasks/wait_for_splunk_instance.yml
```diff
@@ -1,15 +1,12 @@
...
-    url: "{{ cert_prefix }}://{{ splunk_instance_address }}:{{ splunk.svc_port }}/services/server/info?output_mode=json"
+    url: "{{ cert_prefix }}://{{ splunk_instance_address }}:{{ splunk.svc_port }}"
     method: GET
-    user: "{{ splunk.admin_user }}"
-    password: "{{ splunk.password }}"
     validate_certs: false
   register: task_response
   until:
     - task_response.status == 200
-    - lookup('pipe', 'date +"%s"')|int - task_response.json.entry[0].content.startup_time > 10
```
nwang92: cc @arctan5x do you remember what this check was for? I wonder if it's safe to remove now that most of this provisioning is done asynchronously

nwang92:
In general, looks good to me. I wonder if there are any plays or tasks that require knowing a remote instance's admin password? 

If this is the only situation this occurs, then maybe we can close out https://github.com/splunk/splunk-ansible/issues/206?

roles/splunk_common/tasks/wait_for_splunk_instance.yml
arctan5x: Splunk used to restart couple times through Ansible provisioning and wanted to give some time after Splunk starts to stabilize.

mikedickey:
> In general, looks good to me. I wonder if there are any plays or tasks that require knowing a remote instance's admin password?
> 
> If this is the only situation this occurs, then maybe we can close out #206?

This came up via work on breaking down CRDs into smaller components for our K8s operator. So far, I've identified only two cases where component A requires an authentication secret for component B:

* idxc.pass4SymmKey is required for search heads to communicate with cluster masters / indexers
* pass4SymmKey (the top level one) is required by any instance to register with a license master.

This of course excludes the use of pass4SymmKey secrets inside the context of components, like search head cluster members communicating with captain, index cluster members communicating with cluster master, etc.

I haven't encountered any case where an instance has to have the same admin password as any other instance. I believe the pass4SymmKey secrets are always used instead.

Issue #383, opened 2020-02-12
SPLUNK_LICENSE_MASTER_URL uses the protocol & management port from the local Splunk instance


tmartin14:
When setting up a docker instance to use a license master, the SPLUNK_LICENSE_MASTER_URL setting only allows for the hostname or ip address of the server.   The underlying code assumes the protocol and port to be used for the connection are the same as the local instance.  

The current code resides in [https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_common/tasks/set_as_license_slave.yml](https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_common/tasks/set_as_license_slave.yml):

Potential fixes could be either providing separate variables or simply allow the SPLUNK_LICENSE_MASTER_URL to include <protocol>://<dns-or-ip>:<port>.    

Sample invocation: 
```
docker run -t --name=splunk_test -d -p 8000:8000 -p 8088:8088 \
  -e "SPLUNK_START_ARGS=--accept-license" \
  -e "SPLUNK_PASSWORD=<password>" \
  -e "SPLUNK_LICENSE_MASTER_URL=xx.xx.xxxxxxxx.xx" \
  -e "SPLUNK_ROLE=splunk_standalone" \ 
  splunk/splunk:latest
```

nwang92:
Sorry it's taken a while to address this - we'll see if we can squeeze this into the next sprint cycle.

I think we can probably make some adaptability for protocol/scheme and port, ala something like:
```
-e SPLUNK_LICENSE_MASTER_URL=http://license-master:8080
```

nwang92:
The inherent problem of passwords expected to be the same will still exist though - see https://github.com/splunk/splunk-ansible/issues/222 or https://github.com/splunk/splunk-ansible/issues/206.

nwang92:
This will be out in the >=8.0.5 image set

PR #390, opened 2020-02-19
forward servers fix

tod-uma:
There is a bug where the scripts won't add forward_servers when there isn't a s2s config. I've fixed that bug and added molecule tests to make sure forward_servers continues to work in the future.


tod-uma:
This pull request still requires EITHER splunk.s2s_port or splunk.s2s.port be specified in the config. Should we set up something in get_facts to set a default for the s2s port since someone may want to have forward_servers, but not any other s2s configuration?

nwang92:
If I understand correctly, there's a bit of misconception in how we're using `splunk.s2s*`. Part of the plays use `splunk.s2s.port` to setup receiving, but the section modified here also uses `splunk.s2s.port` to setup forwarding.

That's been a problem for us recently and we're looking at approaches to resolve it. For instance: https://github.com/splunk/splunk-ansible/issues/383 - although in this case, we need to know the remote instance's management port (assumed to be 9997) whereas in this case, we need to know the remote instance's s2s port (assumed to be 9997). If you have any ideas, feel free to suggest away!

As far as:
> someone may want to have forward_servers, but not any other s2s configuration
That should be the intent of `splunk.s2s.enable` - if this is true, then we will configure this instance for receiving. Otherwise, it should disable receiving. This currently should not impact forwarding, with the exception being that we're using the same port for both.

tod-uma:
> That's been a problem for us recently and we're looking at approaches to resolve it. For instance: #383 - although in this case, we need to know the remote instance's management port (assumed to be 9997) whereas in this case, we need to know the remote instance's s2s port (assumed to be 9997). If you have any ideas, feel free to suggest away!

I could rewrite the forward-servers logic to require you specify the port for each forward server. I'm not sure how I'd do that and keep backwards compatibility without tons of spaghetti logic though.

nwang92:
Yeah, let's leave that as a separate issue. I'm still wrapping my head around how to make it compatible for both environment variable syntax for the docker-splunk work + default.yml for this work.

PR #393, opened 2020-02-20
fix to incorrect include_tasks causing lint failures

tod-uma:


nwang92:
Ah this was the offending line - I guess this is a new check done that wasn't done by molecule==2.2

PR #394, opened 2020-02-21
Correctly prefix set_search_peers environment variable

alishamayor:
prefixing env variable with SPLUNK_, setting defaults, and adding documentation


pgullette:
I see this has already been merged, but this PR actually introduces a bug that prevents search heads from being peered to cluster master. After #350, the conditionals for these two tasks in roles/splunk_search_head/tasks/main.yml were changed to:

`## Indexer Clustering Scenario
  when:
    - splunk.indexer_cluster
    - splunk.multisite_master is not defined
    - splunk.set_splunk_search_peers is defined
    - splunk.set_splunk_search_peers | bool

## Non Indexer Clustering Scenario
- include_tasks: peer_indexers.yml
  when:
    - not splunk.indexer_cluster
    - splunk.multisite_master is not defined
    - splunk.set_splunk_search_peers is defined
    - splunk.set_splunk_search_peers | bool`

PR #394 fixed some of the logic with the new set_search_peers variable in environ.py, but forgot to update the affected task for search heads as well. Simply updating roles/splunk_search_head/tasks/main.yml to use splunk.set_search_peers instead of splunk.set_splunk_search_peers will fix this issue.

PR #395, opened 2020-02-23
Correcting issue introduced after PR #394

pgullette:
Left a comment on PR #394, but decided to create my own PR to make it even more obvious.

This corrects the issue initially introduced by #350 which was later cleaned up by #394 but left behind this side effect.


alishamayor:
@pgullette Thanks for catching that!

PR #396, opened 2020-02-24
Added license_master_url, cluster_master_url, search_head_captain_url and deployer_url parameters to default.yml

mikedickey:
Added `license_master_url`, `cluster_master_url`, `search_head_captain_url` and `deployer_url` parameters to the `splunk` section of the default.yml config. These may be used to define the location of these components via YAML, without having to use the corresponding `SPLUNK_LICENSE_MASTER_URL`, `SPLUNK_CLUSTER_MASTER_URL`, `SPLUNK_SEARCH_HEAD_CAPTAIN_URL` and `SPLUNK_DEPLOYER_URL`.

Note that the environment variables still work the same as before. This just makes it easier to configure them. Previously, to use a license master, one had to specify the location via an environment variable and the secret (pass4SymmKey) via default.yml. This just makes it a lot easier for users. For example, one can now create a "license_master.yml" config that only includes these parameters and layer it in with other configs:

```
splunk:
  license_master_url: my-license-master
  pass4SymmKey: P@ssw0rd
```

Or, create a "cluster_master.yml" config:

```
splunk:
  cluster_master_url: my-license-master
  idxc:
    pass4SymmKey: P@ssw0rd
```

This PR also includes a change for the standalone instance role so that it works as a standalone search head when a cluster master is defined. Previously, any cluster master config was ignored when running in the standalone role.


inventory/environ.py
```diff
@@ -248,9 +248,11 @@ def getDistributedTopology(vars_scope):
...
     """
-    vars_scope["splunk"]["license_master_included"] = bool(os.environ.get("SPLUNK_LICENSE_MASTER_URL"))
+    vars_scope["splunk"]["license_master_url"] = os.environ.get("SPLUNK_LICENSE_MASTER_URL", vars_scope["splunk"].get("license_master_url"))
+    vars_scope["splunk"]["license_master_included"] = bool(vars_scope["splunk"]["license_master_url"])
     vars_scope["splunk"]["deployer_included"] = bool(os.environ.get("SPLUNK_DEPLOYER_URL"))
-    vars_scope["splunk"]["indexer_cluster"] = bool(os.environ.get("SPLUNK_CLUSTER_MASTER_URL"))
+    vars_scope["splunk"]["cluster_master_url"] = os.environ.get("SPLUNK_CLUSTER_MASTER_URL", vars_scope["splunk"].get("cluster_master_url"))
+    vars_scope["splunk"]["indexer_cluster"] = bool(vars_scope["splunk"].get("cluster_master_url"))
     vars_scope["splunk"]["search_head_cluster"] = bool(os.environ.get("SPLUNK_SEARCH_HEAD_CAPTAIN_URL"))
     vars_scope["splunk"]["search_head_cluster_url"] = os.environ.get("SPLUNK_SEARCH_HEAD_CAPTAIN_URL")
```
nwang92: Should lines 257-256 get the same love by defaulting to `os.environ.get("SPLUNK_SEARCH_HEAD_CAPTAIN_URL", vars_scope["splunk"].get("search_head_cluster_url")`?

inventory/splunk_defaults_linux.yml
```diff
@@ -83,6 +83,8 @@ splunk:
     multisite_replication_factor_total: 3
     multisite_search_factor_origin: 1
     multisite_search_factor_total: 2
+    license_master_url:
+    cluster_master_url:
```
nwang92: Semi-related: we have a `splunk.shc.enable` and `splunk.idxc.enable` which at one point in time may have been used to turn on clustering. But it seems like that is no longer the case and nothing uses those two `enable` flags. Should we consider removing them?

mikedickey: I confirmed that the enable flags were not being used anywhere, and removed references to them in the latest commit. I also moved the calculated booleans out of environ.py and into the plays themselves, which I think is less confusing.

inventory/environ.py
mikedickey: To be consistent, I updated the PR to include support for `splunk.search_head_captain_url` (deprecating the inconsistently-named `splunk.search_head_cluster_url` field which didn't actually work), and also for `splunk.deployer_url`. Like the others, these can be used as alternatives to the SPLUNK_* environment variables, which should still work same as before.

roles/splunk_common/tasks/get_facts.yml
```diff
@@ -74,28 +74,16 @@
...
 
 # determine if we need to set up indexer clusters
-# First, if it is manually specified in the config, use that
 - name: "Setting indexer cluster fact from config"
   set_fact:
-    splunk_indexer_cluster: "{{ splunk.indexer_cluster }}"
-  when:
-    - "'indexer_cluster' in splunk"
+    splunk_indexer_cluster: "true"
+  when: "'cluster_master_url' in splunk and splunk.cluster_master_url"
```
nwang92: Adding a note for mostly myself to revisit + maybe change this later, but I'd prefer facts like this to always be defined and have a default value. That way we don't have to rely on like "is defined" or "is not defined" to drive provisioning logic. Also this could be made into a boolean type instead of a weird string boolean.

nwang92:
Approving this and going to merge, although I'm going to manually kick off docker-splunk tests to make sure all of that still works. If we see any failures there, I'm going to revert these changes.

PR #398, opened 2020-03-02
Reverting standalone changes

nwang92:
Removing the `docker_env`/`is_docker` vars from get_facts.yml because they are not used anywhere. I encountered some errors/test failures when the latest edge image was rolled into our internal tools. Reproduced with:
```
docker run -d -P --name so1 -e SPLUNK_START_ARGS=--accept-license -e SPLUNK_PASSWORD=helloworld -e SPLUNK_STANDALONE_URL=so1 -e DEBUG=true splunk/splunk:edge
```

It seems this change doesn't play very nice with https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_common/tasks/get_facts.yml#L90, and what ends up happening is you end up with some reflexive forwarding. 

@mikedickey I know you added this section in the standalone role from your last PR, but I think we should probably draw some lines around expectations of each role. I'm not sure if it makes sense for standalones to automatically peer the indexer tier. Arguably, this is no longer a "standalone" now and is probably more apt to be considered just a "search head". Also going back to internal usage of these images, there may be some users who create standalones to setup peering to various other roles themselves for the sake of testing.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -88,8 +88,10 @@ def getSplunkInventory(inventory, reName=r"(.*)_URL"):
                 'hosts': list([host.split(':')[0] for host in hosts])
             }
     inventory["all"]["vars"] = getDefaultVars()
+    inventory["all"]["vars"]["docker"] = False
 
     if os.path.isfile("/.dockerenv"):
+        inventory["all"]["vars"]["docker"] = True
         if "localhost" not in inventory["all"]["children"]:
             inventory["all"]["hosts"].append("localhost")
         inventory["_meta"]["hostvars"]["localhost"] = inventory["all"]["vars"]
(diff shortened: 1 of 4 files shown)


mikedickey:
The Operator for K8s currently depends on these changes for the standalone role. Note that they only trigger when a cluster master is defined, either via `splunk.cluster_master_url` or `SPLUNK_CLUSTER_MASTER_URL`. I believe it's incorrect to ignore these parameters, when provided, as having a "standalone" (not clustered) instance that forwards and peers with an indexer cluster is a very common thing. Note there is no change in behavior when these parameters are not defined.

mikedickey:
I think the problem is that `get_facts.yml` is setting `splunk_forward_servers` to itself; see https://github.com/splunk/splunk-ansible/blob/c7f6d6d570bd9852609cbb6b4ffbdfde8b647dec/roles/splunk_common/tasks/get_facts.yml#L106

Those last two plays make over-reaching assumptions, and I think they should both be removed.

nwang92:
Applied a workaround to avoid this case of self-forwarding. When I get some time I'm going to take a closer look at this section to see if there's other ways to capture this logic and not leave variables (like `splunk.forward_servers`) hidden and undocumented.

nwang92:
Going to merge and kick off docker-splunk tests + internal image rebuilds

PR #399, opened 2020-03-04
enabling customizable appserver_port, kvstore_port

qrayven:
Add ability to set custom ports via envs or `default.yml` for services:
- key-value store 
- application server

### DIFF
--- a/inventory/splunk_defaults_windows.yml
+++ b/inventory/splunk_defaults_windows.yml
@@ -55,6 +55,7 @@ splunk:
         password:
         ca:
     appserver_port: 8065
+    kvstore_port: 8191
     http_port: 8000
     http_enableSSL: 0
     http_enableSSL_cert:
--- a/inventory/splunk_defaults_windows.yml
+++ b/inventory/splunk_defaults_windows.yml
@@ -47,8 +47,10 @@ splunk:
         cert:
         password:
         ca:
-    appserver_port: 8065
-    kvstore_port: 8191
+    appserver:
+        port: 8065
+    kvstore:
+        port: 8191
     http_port: 8000
     http_enableSSL: 0
     http_enableSSL_cert:
--- a/inventory/splunk_defaults_windows.yml
+++ b/inventory/splunk_defaults_windows.yml
@@ -27,22 +27,15 @@ splunk:
     allow_upgrade: True
     tar_dir: "splunk"
     opt: &opt "/opt"
-    home: !!python/object/apply:os.path.join &home [*opt, "splunk"]
+    home: &home !!python/object/apply:os.path.join [*opt, "splunk"]
     user: "splunk"
     group: "Administrators"
     exec: !!python/object/apply:os.path.join [*home, "bin", "splunk.exe"]
-    pid:
-        !!python/object/apply:os.path.join [
-            *home,
-            "var",
-            "run",
-            "splunk",
-            "splunkd.pid",
-        ]
+    pid: !!python/object/apply:os.path.join [*home, "var", "run", "splunk", "splunkd.pid"]
     admin_user: "admin"
     root_endpoint:
     password:
-    secret:
+    secret: 
     pass4SymmKey:
     svc_port: 8089
     launch:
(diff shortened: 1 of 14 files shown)


inventory/splunk_defaults_windows.yml
```diff
@@ -46,7 +46,9 @@ splunk:
         ssl: False
         cert:
         password:
-        ca: 
+        ca:
+    appserver_port: 8065
```
nwang92: Nitpicky, but I regret making some of the options here flattened (ex. `http_port`, `http_enableSSL`, etc.). Since these are new additions, I would much prefer to make these nested groups like:
```
splunk:
    kvstore:
        port: 8191
```
to support future maintainability of tuning parameters in specific components of Splunk.

roles/splunk_common/tasks/set_hec_port.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Set HTTP Envent Collector (HEC) Port
+  ini_file:
+    dest: "{{ splunk.home }}/etc/system/local/inputs.conf"
+    section: http
+    option: "port"
+    value: "{{ splunk.hec_port }}"
```
nwang92: Might be possible to do this via CLI in `roles/splunk_common/tasks/set_as_hec_receiver.yml` so we don't need to write conf files. 

https://docs.splunk.com/Documentation/Splunk/8.0.2/Data/UseHECfromtheCLI

roles/splunk_common/tasks/set_appserver_port.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Set appserver port
+  ini_file:
+    dest: "{{ splunk.home }}/etc/system/local/web.conf"
+    section: settings
+    option: "appServerPorts"
+    value: "{{ splunk.appserver_port }}"
```
nwang92: I'd also add 
```
owner: "{{ splunk.user }}"
group: "{{ splunk.group }}"
```
to the ini_file module

inventory/splunk_defaults_windows.yml
qrayven: fixed for `kvstore` and `appstore`

roles/splunk_common/tasks/set_hec_port.yml
qrayven: ok, task removed

roles/splunk_common/tasks/set_appserver_port.yml
qrayven: added for both tasks

roles/splunk_common/tasks/main.yml
```diff
@@ -87,6 +87,21 @@
...
+  when:
+    - "'appserver_port' in splunk"
+    - splunk.appserver_port | int != 8065
+
+- include_tasks: set_kvstore_port.yml
+  when:
+    - "'kvstore_port' in splunk"
+    - splunk.kvstore_port | int != 8191
+
+- include_tasks: set_hec_port.yml
```
nwang92: I believe my PR should take care of the HEC port: https://github.com/splunk/splunk-ansible/pull/402

roles/splunk_common/tasks/main.yml
```diff
@@ -87,6 +87,21 @@
     - "'svc_port' in splunk"
     - splunk.svc_port | int != 8089
 
+- include_tasks: set_appserver_port.yml
+  when:
+    - "'appserver_port' in splunk"
```
nwang92: This variable has been renamed now

roles/splunk_common/tasks/main.yml
```diff
@@ -87,6 +87,21 @@
...
+    - splunk.appserver_port | int != 8065
+
+- include_tasks: set_kvstore_port.yml
+  when:
+    - "'kvstore_port' in splunk"
```
nwang92: This variable has been renamed now

roles/splunk_common/tasks/main.yml
qrayven: This is actually a good catch, After I deleted `set_hec_port.yml`, I forgot to delete the task. Thanks, removed.

roles/splunk_common/tasks/main.yml
qrayven: good catch as well, fixed

roles/splunk_common/tasks/main.yml
qrayven: good catch as well, fixed

PR #401, opened 2020-03-04
DFS Support

jmeixensperger:
Goes along with docker-splunk changes: https://github.com/splunk/docker-splunk/pull/325

Added spark_master and spark_worker roles with ansible plays for spark configuration and execution. Also included logic to install the spark package on search heads that are connected to spark clusters (required for DFS search to work).


roles/splunk_spark_master/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: "Set privilege escalation user"
+  set_fact:
+    privileged_user: "{% if ansible_system is match('CYGWIN*|Win32NT') %}Administrator{% else %}root{% endif %}"
```
nwang92: I don't think you need to define `privileged_user` here - it should already be available because you have `roles/splunk_spark_master/meta/main.yml` with a dependency on `splunk_common`.

library/download_logfile.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/python
```
nwang92: I would recommend finding a different way to do this - this shouldn't be in Github

roles/splunk_common/tasks/s3_download.yml
```diff
@@ -0,0 +1,15 @@
+---
```
nwang92: Same as above, let's find an alternative as I don't want this to be public.

roles/splunk_spark_worker/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: "Set privilege escalation user"
+  set_fact:
+    privileged_user: "{% if ansible_system is match('CYGWIN*|Win32NT') %}Administrator{% else %}root{% endif %}"
```
nwang92: Same as https://github.com/splunk/splunk-ansible/pull/401/files#r390009484

roles/splunk_spark_worker/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
```
nwang92: Do we need this dependency actually? This starts both `splunk` and `spark_worker` processes in the same container, and it does not seem like we are using anything from `splunk`.

roles/splunk_common/tasks/get_spark.yml
```diff
@@ -0,0 +1,39 @@
...
+  file:
+    path: "{{ splunk.dfs.spark_home }}"
+    mode: 0777
+    state: directory
+  become_user: "{{ privileged_user }}"
+  become: yes
+  register: spark_directory
+
+- name: Move Spark package to directory
+  shell: /bin/bash -c "mv /tmp/{{ versioned_name }}/* {{ splunk.dfs.spark_home }}"
```
nwang92: I don't think you need the `/bin/bash -c` here?

roles/splunk_common/tasks/get_spark.yml
```diff
@@ -0,0 +1,39 @@
...
+  when:
+    - spark_directory is defined
+
+- name: Chmod Spark directory
```
nwang92: I'm not sure if you need this task either - this sets the entire `spark_home` dir to be owned by the `splunk:splunk` user/group, but when you create the configurations and start `spark`, it is being run as `root`.

Should we be running `spark` as root? Or do you want to run it as the `splunk` user?

inventory/splunk_defaults_linux.yml
```diff
@@ -70,13 +77,19 @@ splunk:
...
+        spark_home: "/opt/spark"
+        java_home: "/opt/jdk"
         enable: False
-        port: 9000
+        port: 7777
         dfc_num_slots: 4
         dfw_num_slots: 10
         dfw_num_slots_enabled: False
         spark_master_host: "127.0.0.1"
-        spark_master_webui_port: 8080
```
nwang92: I believe changing these might break @mikedickey's work on https://github.com/splunk/splunk-operator/ and https://github.com/splunk/docker-spark with

inventory/splunk_defaults_linux.yml
```diff
@@ -7,6 +7,13 @@ retry_num: 60
 wait_for_splunk_retry_num: 60
 shc_sync_retry_num: 60
 
+openjdk:
```
nwang92: Let's remove this section since it appears we're not using it.

mikedickey:
I don't agree that downloading JAVA and Spark at runtime is a good approach. I also don't believe that our playbooks should have a "spark worker" role, or that there should be two different ways to do these things.

roles/splunk_spark_master/tasks/main.yml
jmeixensperger: You are correct, I added this in before I included the splunk_common dependency and forgot to remove it.

roles/splunk_spark_worker/meta/main.yml
jmeixensperger: DFS team mentioned that splunk is **_not needed_** on the containers, but it is something they requested to capture/analyze spark logs. I think it's ok to remove.

jmeixensperger:
> I don't agree that downloading JAVA and Spark at runtime is a good approach. I also don't believe that our playbooks should have a "spark worker" role, or that there should be two different ways to do these things.

I have baked java/spark into the spark images that I've been using, however the DFS team requires spark to be installed on federated search heads. In their words, the search heads use "some spark libraries to submit the job to the spark master".

roles/splunk_common/tasks/get_spark.yml
jmeixensperger: We didn't see any issues starting spark as the root user, however the spark logs complained about access to files under the /opt/spark directory once the DFS search is executed. DFS team explained that they expected the directory/files to be owned by the same user that is starting splunk, but I'm still not sure why...
I can test starting spark as the splunk user to be consistent, but let's discuss this further.

mikedickey:
> > I don't agree that downloading JAVA and Spark at runtime is a good approach. I also don't believe that our playbooks should have a "spark worker" role, or that there should be two different ways to do these things.
> 
> I have baked java/spark into the spark images that I've been using, however the DFS team requires spark to be installed on federated search heads. In their words, the search heads use "some spark libraries to submit the job to the spark master".

We already have a public and supported spark container image for DFS at https://github.com/splunk/docker-spark (this is posted docker hub as `splunk/spark`). We do not need or want to have another one. I think we need to sync up internally before proceeding any further on this work.

PR #404, opened 2020-03-10
Register jre path to Splunk launch

arctan5x:
Also register jre path to Splunk launch in Redhat. Existing 3 tasks can run in both Debian and Redhat. Verified this works with Splunk 8.0.2 and Redhat8 env.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## Navigation
 
+* [8.0.2.1](#8021)
 * [8.0.2](#802)
 * [8.0.1](#801)
 * [8.0.0](#800)
+* [7.3.4.2](#7342)
 * [7.3.4](#734)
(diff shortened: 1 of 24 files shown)


nwang92:
Actually, you may just want to do this for all platforms? CentOS as well?

nwang92:
Actually, it looks like Java isn't supported on CentOS?

alishamayor:
> Actually, it looks like Java isn't supported on CentOS?

I'm pretty sure you can `yum install java-1.8.0-openjdk-devel`

alishamayor:
(sorry, accidentally hit close and comment)

nwang92:
I wonder if you could also remove the `apt` + `command: microdnf ...` tasks using https://docs.ansible.com/ansible/latest/modules/package_module.html

You probably still need to use conditions on the package name, as it differs from OS-to-OS.

PR #406, opened 2020-03-12
Notifying of restart only when necessary

nwang92:
This is to address https://github.com/splunk/docker-splunk/issues/322

Notifying of restart only when splunktcp has changed and Splunk is running. I'm not super happy with the way this is done, so if anyone has any better suggestions I'm very open to it :P


nwang92:
Looking at test failures

PR #410, opened 2020-03-16
Feature/remote execution

nwang92:
It seems like a lot of members of the community are executing `splunk-ansible` remotely, whereas the `docker-splunk` use case relies on running Ansible through the local connection.

This PR provides some documentation to facilitate the ability to natively run `splunk-ansible` remotely, through some controller node such as Ansible Tower/AWX or some jump host. There are some caveats which I still need to document, but I've been able to get this working for simple cases.

Opening this PR for public visibility and for others on the team/in the community to provide some constructive feedback to see if this is viable for them.


docs/execution_patterns/EXECUTION.md
```diff
@@ -0,0 +1,41 @@
+# Execution
+
+The Ansible plays in the `splunk-ansible` project are intended to be run locally on each instance/host of the Splunk Enterprise deployment. This means that each host asyncronously sets itself up using Ansible roles to form the final desired topology which is most clearly displayed through the [docker-splunk](https://github.com/splunk/docker-splunk) project.
```
bb03, line 3: I would probably rewrite this a little.  If the user just reads the first sentence, it makes it seem like the only route would be to run these would be separately on each host.  I suggest something like this.

The Ansible plays in the `splunk-ansible` project can be run in two ways; separately on each instance/host of the Splunk Enterprise deployment or through more traditional separation of control nodes and managed nodes.  In the first method each host asynchronously sets itself up using Ansible roles to form the final desired topology which is most clearly displayed through the [docker-splunk](https://github.com/splunk/docker-splunk) project.  The more traditional methods are outlined below.

Also, `asyncronously` is misspelled in your original doc

docs/execution_patterns/EXECUTION.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+## Local
+Local connection is the intended mode of using `splunk-ansible`. The dynamic inventory script `environ.py` reads environment variables and maps them into Ansible run-time variables that determine how Splunk Enterprise is setup. 
+
+In order to bring up the most basic Splunk standalone instance on a local host, you can run the following:
+
+```bash
+export SPLUNK_PASSWORD=helloworld
+export SPLUNK_BUILD_URL=https://download.splunk.com/products/splunk/releases/8.0.2.1/linux/splunk-8.0.2.1-f002026bad55-Linux-x86_64.tgz
```
bb03, line 24: Do we want to use 8.0.3?

docs/execution_patterns/remote/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+![diagram](./remote-diagram.png)
+
+## Configure hosts
+First, we'll need a few hosts to run these Ansible plays against. For convenience, we can use a set of containers to mimic a baremetal hosts running `sshd`. These hosts are also setup to pre-install the `splunk` user. For more information, please see the [Dockerfile](./Dockerfile) in used.
+
+Additionally, see the [docker-compose.yml](./docker-compose.yml) file included for how the multiple containers are defined to assume the roles in the Splunk deployment.
+```bash
+$ docker-compose up -d
+```
```
bb03, line 12: I'm not sure if you want to specify it like this.  The way it's written makes it seem more like an example.  Especially when interacting with the swarm you would probably use "docker stack" to deploy rather than docker-compose (which iirc works only on single hosts)

nwang92: My intention was to make this an example, but I can adjust the wording to be more clear. Basically this use case is meant to show how to use `splunk-ansible` from a control node, but IMO it's easier to walk through a more "interactive" demo using containers as target hots instead of showing a static IPs that a user may not be able to repeat.

Issue #413, opened 2020-03-18
splunk_deployment_server role attempts to disable non-existent apps


scalatar:
Maybe I'm missing something but I can't seem to prevent the splunk_deployment_server role from failing during build. 

In splunk_deployment_server/tasks/main.yml the order of tasks is as follows:
1) Enable forwarding (no issues there):
`- include_tasks: ../../../roles/splunk_common/tasks/enable_forwarding.yml`
2) Create deployment apps:
```
- include_tasks: create_deployment_apps.yml
  when:
    - splunk.apps_location
```
3) **During create deployment** > apps are installed, to `splunk.apps_path.default`:
```
- include_tasks: install_deployment_apps.yml
  vars:
    app_url: "{{ item }}"
  loop: "{{ splunk.apps_location }}"
```

4) **During create deployment** > Installed apps are enumerated into `installed_apps`, which includes the apps we just installed:
`- include_tasks: ../../../roles/splunk_common/tasks/find_installed_apps.yml`

5) **During create deployment** > Apps are **copied** from `splunk.apps_path.default` to `splunk.apps_path.deployment` using `installed_apps` variable list:
```
# TODO: Might be better to use synchronize here, but we'll need rsync installed
- name: Copy installed apps to deployment-apps
  command: "cp -R {{ splunk.app_paths.default }}/{{ item }} {{ splunk.app_paths.deployment }}"
  register: copy_apps
  changed_when: copy_apps.rc == 0
  failed_when: copy_apps.rc != 0
  become: yes
  become_user: "{{ splunk.user }}"
  with_items: "{{ installed_apps }}"
```

6) **During create deployment** > Installed apps are removed

```
- name: Remove installed apps
  file:
    path: "{{ splunk.app_paths.default }}/{{ item }}"
    state: absent
  become: yes
  become_user: "{{ splunk.user }}"
  with_items: "{{ installed_apps }}"
  notify:
    - Restart the splunkd service

- name: Flush restart handlers
  meta: flush_handlers
```

7) Now we're all done create deployment and we're back in splunk_deployment_server main and we execute `disable_installed_apps`, which uses the old updated `installed_apps` variable containing the apps we just removed in step 6 above:
```
- include_tasks: ../../../roles/splunk_common/tasks/disable_installed_apps.yml
  when:
    - splunk.apps_location
```
and the problem I think is here, where the already built `installed_apps` is being used
```
- include_tasks: find_installed_apps.yml
  when: not installed_apps

[...]

PR #416, opened 2020-03-19
Cleaning up how apps are copied and disabled

nwang92:
I went down a small rabbit hole with https://github.com/splunk/splunk-ansible/issues/413, but the changes here are the following:
* Deployment server will no longer remove apps in etc/apps - it will just disable them
* Deployment server re-uses `install_apps.yml` playbook in splunk_common
* All copying of apps into other directories will now use a common playbook
* When copying apps, we will *not* copy the `local` dir (this should prevent copying a disabled app)


roles/splunk_common/tasks/copy_installed_apps.yml
```diff
@@ -0,0 +1,11 @@
+---
+# TODO: Might be better to use synchronize here, but we'll need rsync installed
+- name: "Copy installed apps to {{ dest }}"
+  shell:
+    cmd: "set -o pipefail && tar -c --exclude=local {{ item }} | tar -x -C {{ dest }}"
+    chdir: "{{ splunk.app_paths.default }}"
+  become: yes
+  become_user: "{{ splunk.user }}"
+  with_items: "{{ installed_apps }}"
```
hendolim: I believe you mean ```with_items: "{{ apps }}"```, though it doesn't matter much since `installed_apps` is the only var being passed here.

nwang92: True, this should be `{{ apps }}`, I can make the change.

pgullette:
Changing the copy behavior to exclude the local directory addresses my recently open #417 issue. In my case, this fix would work, but since this affects at least deployment server role and cluster master role, is there a danger in excluding the entire local folder since a custom app may have legit items in its local folder that are needed wherever the bundle is pushed?

In my #417, I threw out an idea of uninstalling and reinstalling the app before doing the copy to ensure you still have original app state, but it would ensure you don't also have a disabled app due to the ansible task from a previous playbook run. This approach works as long as the state of the app doesn't change from original app state, and I could definitely see that happening with a deployment server app. I'm not really sure what the cleanest and least disruptive option is.

nwang92:
@pgullette 
From what I know, all published apps we develop internally at Splunk never include a `local` directory (for obvious reasons wrt how Splunk configuration management is done). I'm going to assume most (if not all) apps on Splunkbase also adhere to the same policy. This was the main premise I went off of when deciding to exclude `local` during this new copy step.

> is there a danger in excluding the entire local folder since a custom app may have legit items in its local folder that are needed wherever the bundle is pushed
That's a very valid point. Although if you're maintaining/developing an app, I would assume that a change to some config you want to deploy to your cluster shouldn't be made in `local` in the first place. Any change made in `local` can also be made in `default` after all.

Additionally, copying `local` may create new problems. I'll need to double-check to confirm this behavior, but when pushing a bundle out containing `local` you may be replacing the peer's local. For instance, I'm imagining a case of a bundle containing `local/inputs.conf` with:
```
[tcp://1514]
disabled = 0
```
But on your peer, you have an existing `local/inputs.conf` with:
```
[tcp://1514]
disabled = 0

[tcp://10514]
disabled = 0
```

The end result after the bundle push would be you losing the 10514 TCP input. This may be a bit of a contrived scenario but as a whole: if Splunk Enterprise guarantees that your `local` directory remains untouched, I don't see why extending it to these playbooks would be much different.

pgullette:
I agree with you that not shipping an app to be installed with configs in local is best practice. And I think this may be the way to go forward and potentially the best way. I would say it should be documented in the corresponding md files to be sure that when a person uses custom apps, this playbook makes no guarantee about configs stored in local.

As for your potentially contrived scenario, the splunk docs do indicate that any local changes made in the slave-apps directory directly on the peer will be overwritten with the next bundle push. I can't say for certain if that same behavior applies to an instance managed by a deployment server, but I would expect the bundle push/extract logic to be shared between the various orchestration mechanisms splunk provides.

nwang92:
Sounds good, I'll add documentation in both this repo and docker-splunk in a subsequent PR

PR #425, opened 2020-03-30
Added a bearer token authentication for splunk download

arctan5x:
- SPLUNK_BUILD_URL_BEARER_TOKEN can be used to pull Splunk builds against remote paths that require a bearer token authentication.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -304,9 +304,13 @@ def getSplunkBuild(vars_scope):
     """
     Determine the location of the Splunk build
     """
+    bearer_token_provided = os.environ.get("SPLUNK_BUILD_URL_BEARER_TOKEN")
+    if bearer_token_provided:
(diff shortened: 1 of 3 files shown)


nwang92:
This works, but I think we can optimize the workflow here because there's some redundancy in `install_splunk_tgz.yml`.

I'm thinking:
- If `splunk.build_location` starts with "http", execute the `get_url` task; (optionally include the headers if supplied)
- If `splunk.build_location` starts with "http", dynamically change the `splunk.build_location` to a specified download destination
- Run `unarchive` task on `splunk.build_location` (reuse "Install Splunk (Linux) from network" and remove "Install Splunk (Linux) from network")

Following this, I think it's probably safe to remove `splunk.build_remote_src` because with these changes, the `unarchive` task will always run on a locally-available file.

arctan5x:
Agree with your point Nelson. Will make the necessary change.

Issue #431, opened 2020-04-07
Container image: Support for other runtimes than Docker


dadux:
Hi all,

I'd love to migrate our Splunk cluster to Kubernetes, and was playing with the splunk-operator.

When the container starts, it fails with the following error : 

```ERROR! Specified hosts and/or --limit does not match any hosts``` 

I believe this is due to a missing .dockerenv file :
https://github.com/splunk/splunk-ansible/blob/13ce087937a4cbb6c73822897698cbaf6fc40b55/inventory/environ.py#L93 
which treats container differently with localhost.

The `.dockerenv` is mounted for each containers by the docker daemon only, but in our case we are running `containerd` and it does not mount it. I believe users running `CRI-O` in their clusters will run into the same issue.

How can we not rely on this file ? Adding an environment variable ?

nwang92:
👍 let me take a look

nwang92:
I kicked off another build for the `edge` tag, could you try that and see if it works? The image is `splunk/splunk:edge`

dadux:
Confirmed ! latest `splunk/splunk:edge` fixes the problem.

Thanks for a very speedy resolution !

PR #433, opened 2020-04-07
Checking if deployed in Kubernetes

nwang92:
So reviewing our [docs in docker-splunk](https://splunk.github.io/docker-splunk/), we don't officially support other container runtimes nor do we officially support Kubernetes :P 

Because people are using Kubernetes, we should probably expect that different CRIs may be installed. If Kube works with that CRI, I don't see why these images wouldn't - although at the same time, we're currently only testing with the Docker container engine so I can't anticipate what issues may arise between other runtimes.


nwang92:
In response to https://github.com/splunk/splunk-ansible/issues/431 and https://github.com/splunk/docker-splunk/issues/352

Issue #426, opened 2020-03-31
ability to specify config files with the same name in /tmp/defaults/default.yml


j4ys0n:
it would be handy to be able to specify files of the same name in different paths using the `/tmp/defaults/default.yml` file. 

like:
```yaml
splunk:
    conf:
        inputs:
            directory: /opt/splunk/etc/apps/search/local
            content:
                "udp://8125":
                    connection_host: ip
                    index: docker_metrics
                    sourcetype: statsd
        inputs:
            directory: /opt/splunk/etc/apps/mongodb/local
            content:
                "mongo_admin://top":
                    index: mongodb
                    disabled: 1
                    interval: 15
                    commands: top ping
                    server: mongodb
                    port: 27017
                    username: admin
                    password: admin
```
currently only the last specification is written because it's a duplicate key.

a way to input a list would be awesome!

nwang92:
Closing because the schema for this has been modified to support arrays. We'll republish the `splunk/splunk:8.0.3` image tomorrow for other reasons and we'll include this change.

Issue #375, opened 2020-02-04
When using default.yml to add configuration, empty keys aren't added to config files


mhoogcarspel-splunk:
When using default.yml to add configuration, empty keys aren't added to config files

This can be interesting for example for adding empty serverclasses.

Example:
```
splunk:
	conf:
		serverclass:
			directory: /opt/splunk/etc/system/local
			content:
				global:
					targetRepositoryLocation: $SPLUNK_HOME/etc/apps
					disabled: false
					excludeFromUpdate: "$app_root$/.git*"
					"serverClass:secrets": 
						repositoryLocation: /etc/splunk/secrets
					"serverClass:secrets:app:test": {}
					"serverClass:secrets:app:*": []
					"serverClass:secrets:app:universal_secrets":
						"whitelist.0": "*splunk-ingester*"
```
results in
```
[serverClass:secrets:app:test]
[serverClass:secrets:app:*]
```
missing in /opt/splunk/etc/system/local/serverclass.conf after.

mhoogcarspel-splunk:
Possible workarounds:
Create non-empty keys where possible.

Include files or apps instead of using using the config option in default.yml
https://splunk.github.io/splunk-ansible/ADVANCED.html#app

nwang92:
Not sure if the formatting got messed up, but I assume the default.yml is supposed to look like:
```
splunk:
  conf:
    serverclass:
      directory: /opt/splunk/etc/system/local
      content:
        global:
          targetRepositoryLocation: $SPLUNK_HOME/etc/apps
          disabled: false
          excludeFromUpdate: "$app_root$/.git*"
        "serverClass:secrets":
          repositoryLocation: /etc/splunk/secrets
        "serverClass:secrets:app:test": {}
        "serverClass:secrets:app:*": []
        "serverClass:secrets:app:universal_secrets":
          "whitelist.0": "*splunk-ingester*"
...
```
The `serverClass:secrets:app:test` should be at the same level as global correct? Otherwise, it may not be treating it as a section as intended.

mhoogcarspel-splunk:
I had to re-intend before copy pasting, I made a mistake there, sorry,

In the testing it was all at the same level as this serverclass, and the ones with the whitelist like below, did get rendered:

```
"serverClass:secrets:app:universal_secrets":
          "whitelist.0": "*splunk-ingester*"
```

nwang92:
[...]

PR #442, opened 2020-04-13
Bugfix/support conf array

nwang92:
This should fix + add support for https://github.com/splunk/splunk-ansible/issues/426 and https://github.com/splunk/splunk-ansible/issues/375


nwang92:
Adding some more context behind this PR.

This changes some of the type handling that `splunk.conf` supports. In the currently published form, `splunk.conf` is a dict that looks something like:
```
conf:
    user-prefs:
      directory: /opt/splunk/etc/users/admin/user-prefs/local
      content:
        general:
          default_namespace: appboilerplate
          search_syntax_highlighting: dark
          search_assistant:
        "serverClass:secrets:app:test": {}
```

Maps can't be used to achieve the goal of https://github.com/splunk/splunk-ansible/issues/426, so instead we should be accepting an array. The new schema for this will look like:
```
conf:
  - key: user-prefs
     value:
       directory: /opt/splunk/etc/users/admin/user-prefs/local
       content:
         general:
           default_namespace: appboilerplate
           search_syntax_highlighting: dark
           search_assistant:
         "serverClass:secrets:app:test": {}
```

This isn't the desired schema I'd ordinarily go with, but it actually makes support both dict and array inputs easier which is probably best for backwards compatibility. If `splunk.conf` is of the original dict type, we can use Ansible's `dict2items` filter to convert it to the aforementioned array. 

I also added support for empty stanzas as seen with `"serverClass:secrets:app:test": {}` which should address https://github.com/splunk/splunk-ansible/issues/375

PR #451, opened 2020-04-16
Adding special handling for ESS

nwang92:
Apparently we can't exclude the `local` dir when installing ESS through the deployer. This is mainly the way ESS is packaged, which requires the use of a Splunk search command to actually extract all the dependencies and finish the installation.


nwang92:
There is a nice `--deployment_type shc_deployer` parameter with ESS >= 6.0.0: https://docs.splunk.com/Documentation/ES/6.0.0/Install/InstallEnterpriseSecuritySHC#Installing_Splunk_Enterprise_Security_from_the_command_line_in_a_search_head_cluster_environment

But this doesn't work for older versions of ESS - it would be nice to eventually move to this if possible

Issue #437, opened 2020-04-10
outputs.conf must be created before ftr


mikedickey:
Submitted by @rfaircloth-splunk 

Splunk indexes events immediately on startup this is causing events that should be on the indexer to be found on the local buckets of the SH in the SHC and CM/LM outputs.conf must be generated before splunk is started

Issue #232, opened 2019-08-02
Outputs configuration should not use rest


rfaircloth-splunk:
Using rest causes splunk to index locally until forwarding is enabled this is an incorrect behavior

nwang92:
We opted for the REST API over writing conf files for the sake of backwards-compatibility + future-proofing the spec. If anything changes when writing conf files directly, there's a chance btool prevents Splunk from starting - whereas we *should* be able to expect the REST API to be consistent and handle any changes when necessary.

There's also some added benefit of REST API changes being reflected immediately vs. files on disk changes requiring a full Splunk restart. In the end, I suppose there will be a case of splunk-indexing-locally-until-forwarding-is-enabled, but this period shouldn't last very long and it comes with the tradeoff of increased start-up time.

PR #453, opened 2020-04-20
Change the way forwarding config is generated

hendolim:
As mentioned in #437 and #232, outputs.conf is currrently generated using REST api. This pose a problem since it means that the conf file cannot be generated until splunk is started. As a result, data is indexed locally as soon as splunk starts due to the lack of config that disables it.

The changes here:
- Every stanzas in outputs.conf file is written manually instead of using splunkd API.
- Move enable_forwarding task from each role to splunk_common tasks before splunk is started.


roles/splunk_common/tasks/add_forward_server.yml
```diff
@@ -17,24 +17,21 @@
...
+    value: "{{ item.value }}"
+  with_items:
+    - {key: "server", value: "{% for fwd in forward_servers %}{{ fwd }}:{{ splunk.s2s.port }}{{ ',' if not loop.last else '' }}{% endfor %}"}
+    - {key: "clientCert", value: "{{ splunk.s2s.cert if splunk.s2s is defined and splunk.s2s.cert is defined else ''}}"}
+    - {key: "sslPassword", value: "{{ splunk.s2s.password if splunk.s2s is defined and splunk.s2s.password is defined else ''}}"}
+    - {key: "useClientSSLCompression", value: "{{ 'true' if splunk.s2s is defined and splunk.s2s.ssl is defined and splunk.s2s.ssl else '' }}"}
+  when:
+    - '"s2s" in splunk and "ssl" in splunk.s2s and splunk.s2s.ssl | bool'
+    - item.value | length > 0
   notify:
```
nwang92, line 35: Just a note here on idempotency: I don't believe we can use `notify` to trigger Splunk to restart. I believe what happens here is when this play runs for the first time and creates the outputs.conf file, the status will be "changed" and the "restart Splunk" handler will get triggered. But at the time of execution, Splunk isn't running - shortly after we call the Splunk start playbooks/tasks. But the notify handler doesn't get cleared, so at the end of the site.yml Splunk will restart when it doesn't need to.

At the same time, we should leave in some mechanism for triggering restarts too because some people may be using these playbooks outside of the context of docker-splunk. You may be calling `site.yml` on an already-running Splunk instance, in which case you will want to trigger the handler.

I haven't found a good way of handling this other than what I did here https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_common/tasks/s2s/configure_splunktcp.yml

hendolim: Okay, pushing a change to include the trigger you have instead

nwang92:
Approving for now, although I guess we'll need to see how the docker-splunk + our internal tests look with this change

nwang92:
Ok I think this broke a few of the distributed tests: https://app.circleci.com/pipelines/github/splunk/docker-splunk/721/workflows/4cfc1aff-83a5-4b47-a0e8-bcffe4a5368d/jobs/1943/steps

Let me figure out what went wrong

nwang92:
Got it, it seems like the indexers might not be indexing?
```
[ansible@idx1 splunk]$ cd /opt/splunk/etc/system/local/
[ansible@idx1 local]$ ls
README	inputs.conf  migration.conf  outputs.conf  server.conf
[ansible@idx1 local]$ sudo -u splunk bash
[splunk@idx1 local]$ cat outputs.conf

[indexAndForward]
index = false
```

hendolim:
sorry, just saw this now @nwang92 . Seems like you already got this covered on https://github.com/splunk/splunk-ansible/pull/460. Thanks 👍

PR #457, opened 2020-04-27
Upgrading molecule to version 3

nwang92:
Molecule 3.x changing how the `lint` directive is used. I'm removing it entirely because we're already using `ansible-lint` as a separate step/stage of the CircleCI pipeline.

I'm not entirely sure yamllint and flake8 linting are necessary either, but we can add those back in at a later time.


nwang92:
Hmm I seem to have messed up the CI even more :) Will work off this PR to figure out what went wrong

nwang92:
Looked into the errors on the `splunk-ansible-test` job of CircleCI. Based on the exception:
```
TypeError: Invalid special arguments:

  'env': value u'/home/circleci/project/roles/splunk_standalone/.env.yml' of env key 'MOLECULE_ENV_FILE' must be a str
  'env': value u'/home/circleci/project/roles/splunk_standalone/molecule/default/tests' of env key 'MOLECULE_VERIFIER_TEST_DIRECTORY' must be a str
  'env': value u'/home/circleci/project/roles/splunk_standalone/molecule/default' of env key 'MOLECULE_SCENARIO_DIRECTORY' must be a str
 ```
There seems to be some unhandled typing within how molecule is parsing things. Specifically, the values are unicode while molecule expects a string. 

I tried running this using python3 and I was able to get it to run, which leads me to suspect that the molecule project has probably abandoned python2 due to its deprecation. The molecule page on the PyPi repository https://pypi.org/project/molecule/ also suggests that it now requires python3:
```
Requires: Python >=3.6
```

nwang92:
Nevermind - turns out it was the `sh` library: https://github.com/amoffat/sh/releases

The 1.13.0 package was published 12h ago and that's what broke.

nwang92:
Also for posterity: I followed this checklist to upgrade from molecule 2.x to 3.x https://github.com/ansible-community/molecule/issues/2560

PR #462, opened 2020-04-30
Qrayven add custom app server port

bb03:
Made a little mistake when updating this PR https://github.com/splunk/splunk-ansible/pull/399/files

But screw it, going full speed ahead on this one.


docs/advanced/default.yml.spec.md
```diff
@@ -283,6 +283,14 @@ splunk:
   * Determine the port used for Splunk management/remote API calls
   * Default: 8089
 
+  appserver_port: <int>
```
nwang92, line 286: Should be:
```
appserver:
    port: <int>
```
here (and similar for below)

PR #465, opened 2020-05-05
Test/update molecule tests

nwang92:
Updating some of the tests here to leverage molecule more including: multiple platforms, more assertions in the testinfra/pytest bits, and more structured scenarios.

In `splunk_standalone`, the `default` scenario should be all the presets of a standalone instance which include a minimally modified default.yml. The `snowflake` scenario is a heavily modified default.yml with all sorts of options (ex. diff port mappings, custom confs, user-defined secrets, etc).


roles/splunk_standalone/molecule/snowflake/tests/test_snowflake.py
```diff
@@ -0,0 +1,145 @@
...
+
+def test_inputs_conf(host):
+    f = host.file("{}/etc/system/local/inputs.conf".format(SPLUNK_HOME))
+    assert f.exists
+    assert f.user == SPLUNK_USER
+    assert f.group == SPLUNK_GROUP
+    assert f.contains("[splunktcp://9999]")
+    assert f.contains("disabled = 0")
+
+def test_splunk_ports(host):
```
bb03, line 89: Are we intending to test the different default ports - basically are we anticipating that these will ever change or that we'll want to use this for testing custom ports?

nwang92: Yeah, I'm purposely testing the ability to set different ports via the different options in the variables here: roles/splunk_standalone/molecule/snowflake/molecule.yml

roles/splunk_common/tasks/set_root_endpoint.yml
```diff
@@ -1,19 +1,9 @@
...
-    body_format: "form-urlencoded"
-    status_code: 200
-    timeout: 10
-  when: splunk.root_endpoint
-  register: set_root_endpoint
-  changed_when: set_root_endpoint.status == 200
-  no_log: "{{ hide_password }}"
-  notify:
-    - Restart the splunkd service
+  ini_file:
```
jrigassio-splunk, line 3: awesome. much cleaner

PR #467, opened 2020-05-05
Bugfix/java defaults yml

nwang92:
Requested by Matt Modestino, but I'm adding support for java variables so that java can be installed using a default.yml in addition to env vars.

For instance, using the following:
```
# default.yml
java_download_url: https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz
java_update_version: 11.0.2
java_version: openjdk:11
splunk:
    ...
```
should get you openjdk:11 in the container.

Also revamping splunk_heavy_forwarder tests, and adding a test for the above example in the splunk_heavy_forwarder role.


roles/splunk_heavy_forwarder/molecule/default/converge.yml
```diff
@@ -1,82 +1,5 @@
 ---
 - name: Converge
   hosts: all
-  vars:
```
jrigassio-splunk, line 4: Whats the motivation behind moving these vars over to `/molecule/default/molecule.yml`?

roles/splunk_heavy_forwarder/molecule/java/molecule.yml
```diff
@@ -0,0 +1,133 @@
...
+      - "/opt/splunk/var"
+  - name: splunk-hf-redhat8
+    image: geerlingguy/docker-ubi8-ansible:latest
+    pre_build_image: true
+    command: "/lib/systemd/systemd"
+    volumes:
+      - "/opt/splunk/var"
+provisioner:
+  name: ansible
+  inventory:
```
jrigassio-splunk, line 27: I see the vars here again. Does it not make sense for this file and `roles/splunk_heavy_forwarder/molecule/default/molecule.yml` to share most defaults and just overwrite variables specific to the playbook, like 

`java_download_url: https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz`

jrigassio-splunk: or is it best to over-enumerate in ansible/our open-source work?

roles/splunk_heavy_forwarder/molecule/default/converge.yml
nwang92: @arctan5x noticed a while ago that when variables were overwritten in any of the plays, there were some negative consequences. For instance, if some playbook had:
```
- set_fact:
    splunk:
        build_location: /path/to/new/splunk.tgz
```
what ended up happening was the entire `splunk` dict got replaced so that it becomes:
```
{"splunk": {"build_location": "/path/to/new/splunk.tgz"}}
```

This mean things like `splunk.home`, `splunk.user` were undefined afterwards. This only happens during the `molecule` execution, so we suspect it had to do with Ansible's [variable order-of-precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable).

I haven't confirmed if the var movement here fixes this problem, but I was hoping to address that issue Tony found with the test refactoring work.

roles/splunk_heavy_forwarder/molecule/java/molecule.yml
nwang92: Good question, I didn't see an immediate solution from https://molecule.readthedocs.io/en/latest/configuration.html#id12, but I also admittedly didn't look too hard into it either. Let me experiment with a few options

nwang92: I think I have a workaround for this but I'm going to merge this in and see if I can apply that in another PR

PR #470, opened 2020-05-08
Test/update more scenarios

nwang92:
Updating the splunk_heavy_forwarder/manual-forwards scenario and splunk_standalone/systemd scenario to match the new format.


roles/splunk_standalone/molecule/snowflake/group_vars/all.yml
```diff
@@ -1,6 +1,10 @@
 ---
 splunk:
   role: splunk_standalone
+  appserver:
+    port: 8066
+  kvstore:
+    port: 18191
```
alishamayor, line 7: Is this intentionally 18191 instead of 8191?

nwang92: Yeah, the `snowflake` scenario tests the ability to change the kvstore port and makes sure that the change is reflected: https://github.com/splunk/splunk-ansible/pull/470/files#diff-f6dc38764b2c3623237a3a6e7c0e5734R106

PR #471, opened 2020-05-08
Support for earlier splunk versions w/ aws + windows

jmeixensperger:
Tested w/ splunk 7.2.0 and 7.3.0 msi packages on AWS windows2016 EC2 images. The playbooks were consistently failing when attempting to restart splunk after config changes. Ultimately, it timed out because it could not read server.conf (same issue with inputs.conf but this didn't block the restart). With new file permissions, I was able to setup shc and idc w/ both versions.


jmeixensperger:
I need to find an alternate fix that passes idempotency, WIP for now

roles/splunk_common/tasks/set_kvstore_port.yml
```diff
@@ -8,3 +8,9 @@
     owner: "{{ splunk.user }}"
     group: "{{ splunk.group }}"
 
+- name: Adjust server.conf permissions (Windows)
```
nwang92, line 11: This might need to be a separate play because ideally you'll always want this to touch server.conf before Splunk starts, but there's no guarantee `set_kvstore_port.yml` will continue to be the last play included before the start operation.

jmeixensperger:
Ready to merge

PR #475, opened 2020-05-12
Updating splunkdssl play to not write

nwang92:


roles/splunk_common/tasks/enable_splunkd_ssl.yml
```diff
@@ -1,12 +1,14 @@
...
 - name: Enable Splunkd SSL
   ini_file:
     dest: "{{ splunk.home }}/etc/system/local/server.conf"
     section: sslConfig
     option: enableSplunkdSSL
-    value: "{{ splunk.ssl.enable | int }}"
+    value: "false"
     owner: "{{ splunk.user }}"
     group: "{{ splunk.group }}"
+    state: "{{ 'absent' if splunk.ssl.enable else 'present' }}"
```
arctan5x, line 11: if the value is fixed to false when the state is present, would it work as expected?

nwang92: If `splunk.ssl.enable` is set to false, that would indicate that you want splunkd to respond on 8089 over http. So this should correspond to `enableSplunkdSSL = false` which disables SSL. 

I can write a test for this in my docker-splunk branch though.

Issue #438, opened 2020-04-10
Documentation error for default.yml.spec


jenipherc:
It appears to me that 

`splunk:
    upgrade: `

doesn't exist according to 
https://github.com/splunk/splunk-ansible/blob/develop/inventory/environ.py#L414

but it looks like it's mentioned here?
https://github.com/splunk/splunk-ansible/blob/develop/inventory/environ.py#L643

Issue #435, opened 2020-04-09
Missing documentation for some configuration parameters in default.yml


jenipherc:
The following list is generated based on the default.yml from wrapper example
`docker run -it splunk/splunk:8.0.3 create-defaults`

- splunk.multisite_*
- root_endpoint
- service_name
- retry-delay
- wait_for_splunk_retry_num
- splunk.allow_upgrade
- docker
- java_*
- dmc_*
- ansible_connection
- ansible_environment
- splunk.build_url_bearer_token
- splunk.set_search_peers

alishamayor:
Thanks for the heads up! Will take care of it in an upcoming docs update 🤙

PR #482, opened 2020-06-01
Large updates to default.yml spec and keeping it current

nwang92:
Addressing #438


nwang92:
I think this also addresses https://github.com/splunk/splunk-ansible/issues/435

PR #498, opened 2020-06-15
Bugfix/remote lic master

### LINKED ISSUE #383: SPLUNK_LICENSE_MASTER_URL uses the protocol & management port from the local Splunk instance

nwang92:
Fixes #383 

Previously, `splunk.license_master_url` would effectively represent the hostname of the remote Splunk license master, ex: `licmaster1`. We would auto-fill the scheme/port based on the defaults in this project.

I've changed it now so that `splunk.license_master_url` should now include scheme + port after parsing the input `SPLUNK_LICENSE_MASTER_URL`. Theoretically we can extend this to the cluster_master_url, search_head_captain_url, etc. but I don't think that's needed at the moment.

In the new `parseUrl()` function of `environ.py`, I also added some hooks to extract the auth information. I'm hoping that we can eventually address #206 with this by accepting something like: `SPLUNK_LICENSE_MASTER_URL=https://admin:password@licmaster:8089`


inventory/environ.py
```diff
@@ -461,6 +462,38 @@ def getRandomString():
     char_set = string.ascii_uppercase + string.digits
     return ''.join(random.sample(char_set * 6, 6))
 
+def parseUrl(url, vars_scope):
```
nwang92, line 465: I attempted to use `from urlparse import urlparse` as well as `urllib3.utils` to parse this out, but I wasn't getting the correct hostname in certain cases. For instance:
```python
>> urlparse("lm1")
>> ParseResult(scheme='', netloc='', path='lm1', params='', query='', fragment='')
```
Would prefer not to write/maintain our own URL parser here, but IDK if there's anything else folks would recommend.

nwang92:
cc @mikedickey in case this might affect anything on the operator side. I imagine it wouldn't since the interface is the same at least from the Docker env var front.

inventory/environ.py
hendolim: yeah, `urlparse` and the likes only work for valid urls (complete with scheme and domain or at least `//` before the domain). So maybe we can use `urlparse` and then extract the items from the result only if `netscope` != '' , otherwise the entire url is the domain, and get the other items from `vars_scope`. We still need to write codes for this validation and whatnot, though.

bb03: Yeah, I agree with @hendolim.  As long as it results in an overall decrease in lines of code we'll need to maintain.

nwang92: I followed your suggestions and cleaned this section up a bit while keeping the unit tests the same :P It's definitely an improvement I think!

mikedickey:
It looks ok to me. I would like to request that you test it using splunk-operator before merging, if possible.

nwang92:
> It looks ok to me. I would like to request that you test it using splunk-operator before merging, if possible.

I tested this with custom image using the operator, this still seems to work correctly:
```
---
apiVersion: enterprise.splunk.com/v1alpha2
kind: LicenseMaster
metadata:
  name: lm
  namespace: default
  finalizers:
  - enterprise.splunk.com/delete-pvc
spec:
  image: repo.splunk.com/nwang/docker-splunk:remotelm
  imagePullPolicy: Always
  licenseUrl: http://go.sv.splunk.com/splunk-nfr-license
  etcStorage: 1Gi
  varStorage: 5Gi

---
apiVersion: enterprise.splunk.com/v1alpha2
kind: Standalone
metadata:
  name: so
  namespace: default
  finalizers:
  - enterprise.splunk.com/delete-pvc
spec:
  replicas: 1
  image: repo.splunk.com/nwang/docker-splunk:remotelm
  imagePullPolicy: Always
  etcStorage: 1Gi
  varStorage: 5Gi
  licenseMasterRef:
    name: lm
    namespace: default
```

PR #500, opened 2020-06-18
Add ess enablement

bb03:


inventory/environ.py
```diff
@@ -409,6 +410,23 @@ def getHEC(vars_scope):
     else:
         vars_scope["splunk"]["hec"]["ssl"] = bool(vars_scope["splunk"]["hec"].get("ssl"))
 
+def getESSplunkVariables(vars_scope):
+    """
+    Get any special Enterprise Security configuration variables
+    """
+    ssl_enablement_env = os.environ.get("SPLUNK_ES_SSL_ENABLEMENT")
+    if not ssl_enablement_env and (not "es" in vars_scope["splunk"] or not "ssl_enablement" in vars_scope["splunk"]["es"]):
```
kyanite, line 418: True and False
False and False
==================
never goto block of "if condition"

PR #507, opened 2020-07-09
Fixing retry logic around http/s test

nwang92:


nwang92:
For some background: when the 8089 mgmt port responds over HTTP, the status checks don't work because we end up sending an HTTPS request to HTTP and this results in an error. I'm including this error condition as part of the termination of the retry loop, otherwise we end up waiting forever.

PR #519, opened 2020-07-24
INFRA 24441: add dmc_clusterlabel POST calls to DMC in splunk-ansible

TheCamNelson:


roles/splunk_monitor/tasks/post_calls.yml
```diff
@@ -25,4 +25,16 @@
...
\ No newline at end of file
+    - "{{ dmc_group_shc_deployer }}"
+
+- name: Cluster Label POST Requests
+  uri:
+    url: "{{ cert_prefix }}://127.0.0.1:{{ splunk.svc_port }}/services/search/distributed/groups"
+    method: POST
+    user: "{{ splunk.admin_user }}"
+    password: "{{ splunk.password }}"
+    body: "member=localhost:localhost&default=false&name=dmc_indexerclustergroup_{{ item }}"
```
nwang92, line 36: This doesn't feel correct to me - `localhost`/`splunk_monitor` here shouldn't be a member of the indexer cluster group? I would expect that the indexers, maybe the cluster master, form the group?

TheCamNelson: updated

PR #521, opened 2020-07-27
Feature/fix new asan builds

sarahotis:
In enable_asan.yml:
• Adds Debian 10 syntax for future support of that platform (not currently available. Just CentOS for now)
• Adds conditionals for dynamic linking of ASan depending on whether it's an old or new Splunk build (using ASan v2 or v5, respectively). It is possible that, in the future, static linking will be able to replace this dynamic linking of ASan v5. ASan v2 linking will remain the same for backward compatibility.
• Open to any suggestions for more efficient Ansible run.

In Setup docs:
• Added instructions for adding important variables to default.yml (copied from remote execution instructions)


ansible.cfg
```diff
@@ -21,10 +21,6 @@ action_warnings = False
 allow_world_readable_tmpfiles = True
 stderr_callback=debug
 stdout_callback=debug
-localhost_warning = False
-
-[inventory]
-unparsed_is_failed = True
```
nwang92, line 27: @sarahotis and I talked and found this change I made in a previous PR broke some other execution patterns. Will look into this more at a later date.

PR #526, opened 2020-08-05
Adding extra stop condition on monitor; adding license check flags on…

nwang92:
The `{{ splunk.exec }} add forward-server` command can fail under a specific scenario where it's not a first-time run + Splunk has not started. This causes a prompt for the EULA acceptance, which isn't captured here. I'm wondering if we should just revert to generating the config ourselves, but that may be a future nice-to-have.

Adding an extra termination condition when the monitor adds search peers.


bb03:
This is probably a better solution than just generating the config ourselves, because the flags will just be ignored if they aren't needed (vs a potential bug introduced when copying configs)

PR #538, opened 2020-09-10
Remove clustering stanza for idxc and shc

akondur:
Proposing a change to clean out the [clustering] & [shclustering] stanzas as a part of the role splunk_common's tasks. The indexer clustering and search head clustering role specific tasks will take care of populating these stanzas.

The change is required in situations when the container is restarted with persistent volumes. The clean start of a container is prevented as the Splunk instance on the pod uses the clustering stanzas(which persisted from previous runs) instead of using the newly read ansible variables from disk/env variables.

### DIFF
--- a/roles/splunk_common/tasks/main.yml
+++ b/roles/splunk_common/tasks/main.yml
@@ -122,6 +122,10 @@
 - include_tasks: enable_dsp.yml
   when: "'dsp' in splunk and 'enable' in splunk.dsp and splunk.dsp.enable"
 
+- include_tasks: remove_clustering.yml
+  when:
+    - splunk_indexer_cluster or splunk.role == "splunk_search_head"
+
 - include_tasks: start_splunk.yml
 
 - include_tasks: set_certificate_prefix.yml
--- a/roles/splunk_common/tasks/remove_clustering.yml
+++ b/roles/splunk_common/tasks/remove_clustering.yml
@@ -1,0 +1,18 @@
+---
+- name: Remove indexer clustering stanza in order to reconfigure
+  ini_file:
+    dest: "{{ splunk.home }}/etc/system/local/server.conf"
+    section: "clustering"
+    state: absent
+    owner: "{{ splunk.user }}"
+  become: yes
+  become_user: "{{ splunk.user }}"
+
+- name: Remove search head clustering stanza in order to reconfigure
+  ini_file:
+    dest: "{{ splunk.home }}/etc/system/local/server.conf"
+    section: "shclustering"
+    state: absent
+    owner: "{{ splunk.user }}"
+  become: yes
+  become_user: "{{ splunk.user }}"


nwang92:
Could you share a little bit more information about the problem/bug this solves?

akondur:
> Could you share a little bit more information about the problem/bug this solves?

Provided a description

roles/splunk_common/tasks/remove_clustering.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Remove indexer clustering stanza in order to reconfigure
+  ini_file:
+    dest: "{{ splunk.home }}/etc/system/local/server.conf"
```
gaurav-splunk, line 4: So this is assuming that the only place where we have clustering stanza is in etc/system/local/server.conf and not at any other place like etc/slave-apps/ ?

roles/splunk_common/tasks/remove_clustering.yml
```diff
@@ -0,0 +1,18 @@
...
+    section: "clustering"
+    state: absent
+    owner: "{{ splunk.user }}"
+  become: yes
+  become_user: "{{ splunk.user }}"
+
+- name: Remove search head clustering stanza in order to reconfigure
```
nwang92, line 11: Looks like this gets triggered when `splunk_indexer_cluster` which implies there's a cluster master/indexer cluster. But that doesn't necessarily mean there's also a SHC.

roles/splunk_common/tasks/main.yml
```diff
@@ -122,6 +122,10 @@
 - include_tasks: enable_dsp.yml
   when: "'dsp' in splunk and 'enable' in splunk.dsp and splunk.dsp.enable"
 
+- include_tasks: remove_clustering.yml
+  when:
+    - splunk_indexer_cluster or splunk.role == "splunk_search_head"
```
nwang92, line 127: Do you need this for `splunk.role == splunk_indexer` as well?

roles/splunk_common/tasks/remove_clustering.yml
akondur: Yes, it will clear the stanzas only if they exist. It is a no-op if they don't exist.

roles/splunk_common/tasks/main.yml
akondur: I thought splunk_indexer_cluster would take care of that?

nwang92:
If I'm understanding the problem correctly, this only seems necessary if the target indexer cluster or SHC changes. I'm imagining that a `role=splunk_search_head` container gets created and was originally part of indexer cluster of `idxc_label=a` but now you want to reconfigure it to join a different indexer cluster of `shc_label=b`.

However, if it remains the same I'm not sure if we should remove these stanzas. By removing the stanzas and starting Splunk, would there be any data loss? I imagine there's some time it takes to re-join the respective cluster. Also, would this break anything on the captain/cluster master side in terms of stale inventory entries?

roles/splunk_common/tasks/remove_clustering.yml
akondur: Yes we are dealing with the etc/server/local.conf only with this change.

nwang92 (approved):

PR #543, opened 2020-09-16
bugfix/support-upgrades-different-build-hashes

sarahotis:
* Fixes issue in which the build is not upgraded if it is the same version build of a different hash
* Improves setup doc's default.yml instructions

### DIFF
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -54,7 +54,6 @@ Alternatively, you can download the example `default.yml` supplied [here](advanc
 3. Define a few key variables in your `default.yml`:
 * `splunk.role`: the role this instance will play in the Splunk Enterprise deployment. (e.g. `splunk_standalone`)
 * `splunk.build_location`: URL to dynamically fetch the Splunk Enterprise build and install it at run time
-* `splunk.build_remote_src`: this wll be `true` when `splunk.build_location` above is a URL (e.g. `http://`, `https://`, or `file://`), and `false` if it is a path. Note: if a path is provided, the tgz, msi, deb, or rpm file that it points to must be in the container.
 * `splunk.password`: default `admin` user password that Splunk will be provisioned with on first-time run
 
 4. Inspect your newly-created `default.yml` and tweak options as you see fit. For a full list of parameters, please see the [`default.yml.spec`](advanced/default.yml.spec.md#spec).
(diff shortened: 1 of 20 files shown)


roles/splunk_common/tasks/get_facts.yml
```diff
@@ -70,13 +70,16 @@
 - name: "Set current version fact"
   set_fact:
     splunk_current_version: "{{ manifests.files[0].path | regex_search(regexp, '\\1') if (manifests.matched == 1) else '0' }}"
+    splunk_current_build_hash: "{{manifests.files[0].path | regex_search(regexp, '\\3') if (manifests.matched == 1) else '0'}}"
+    splunk_target_build_hash: "{{ splunk.build_location | regex_search(regexp, '\\3') | default('0') }}"
   vars:
-    regexp: 'splunk\D*?-((\d+)\.(\d+)\.(\d+))'
+    regexp: 'splunk\D*?-(\d+\.\d+\.\d+(\.\d+)?)-(.*)-.*-.*.[tgz|msi|rpm|deb]'
```
sarahotis, line 76: stops catching unnecessary matches and also allows for .patch versions
match 1 is version, match 3 is hash

bb03: You may want to see if you can make these matches lazy.  It would probably be okay this way, but .*? should be more efficient

docs/SETUP.md
```diff
@@ -52,9 +52,9 @@ $ docker run -it splunk/splunk:latest create-defaults > default.yml
 Alternatively, you can download the example `default.yml` supplied [here](advanced/default.yml.spec.md#sample).
 
 3. Define a few key variables in your `default.yml`:
-* `splunk.role`: the role this instance will play in the Splunk Enterprise deployment
+* `splunk.role`: the role this instance will play in the Splunk Enterprise deployment. (e.g. `splunk_standalone`)
 * `splunk.build_location`: URL to dynamically fetch the Splunk Enterprise build and install it at run time
-* `splunk.build_remote_src`: this wll be `true` when `splunk.build_location` above is a URL
+* `splunk.build_remote_src`: this wll be `true` when `splunk.build_location` above is a URL (e.g. `http://`, `https://`, or `file://`), and `false` if it is a path. Note: if a path is provided, the tgz, msi, deb, or rpm file that it points to must be in the container.
```
alishamayor, line 57: super NIT, but you're missing an `i` in `will`. Also I think we discussed last time about changing the wording for `build_location` to path since the source could be a local path.

nwang92 (approved):

PR #546, opened 2020-09-22
Increase the timeouts on API calls for Splunk HEC configuration

romain-bellanger:
Hello,

I got some problems on different clustered environments in Kubernetes, running on different cloud infrastructure providers. Some indexer pods would not start because the ansible playbook was failing on this 10s timeout.

Our indexers are connecting to a license master. This is configured just before the HEC (at the end of the splunk_common role), and triggers a restart of the indexers. I assume that Splunk performs some tasks at restart (e.g. fsck) possibly causing the API to respond very slowly. One case was for instance occurred for instance after a complete Splunk site was shut down for infrastructure maintenance. After this operation, the API call was always reaching the timeout, causing playbook and pod failure (the indexers successfully restarted after this and reconnected to the master, but the readinessProbe of the pods depends on playbook success). After patching the timeout, they restarted successfully, with a response time of 18s on one instance. Other cases didn't have a so obvious cause for slow response time after restart.

I only hit this problem on the first task ```Setup global HEC```, but I thought that it would be safer to increase the timeout of all these API calls, to increase the chances that indexers successfully recover after heavy maintenance operations.

### DIFF
--- a/roles/splunk_common/tasks/set_as_hec_receiver.yml
+++ b/roles/splunk_common/tasks/set_as_hec_receiver.yml
@@ -16,7 +16,7 @@
       sslPassword: "{% if 'hec' in splunk and 'password' in splunk.hec and splunk.hec.password %}{{ splunk.hec.password }}{% endif %}"
     body_format: "form-urlencoded"
     status_code: 200
-    timeout: 10
+    timeout: 60
   no_log: "{{ hide_password }}"
 
 - name: Get existing HEC token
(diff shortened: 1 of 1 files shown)


romain-bellanger:
@alishamayor @nwang92 @mikedickey @gaurav-splunk sorry to disturb you, but could you please review this PR? This is an actual problem which we experienced on different clusters. I would love to see it solved upstream, to remove some patching we do on the Splunk docker images, which might break at some point if the code changes. The fix is pretty simple, do you think this could have any unexpected impact? If the timeout expires, the playbook fails anyway.

nwang92 (approved):

nwang92:
Thanks for the contribution @romain-bellanger! Will merge it now

Issue #545, opened 2020-09-21
Address some limitations of the app installation solution


romain-bellanger:
Hello,

While maintaining some clusters in Kubernetes using the alpha version of the [splunk-operator](https://github.com/splunk/splunk-operator/), one of the main issues which are facing is related to the deployment of Splunk applications. I will open a dedicated ticket in the repository of the operator for problems specifically related to the operator, and focus here on the app installation solution provided by splunk-ansible.

## Context

In our traditional environment, we have a lot of applications already maintained in git repositories, automatically validated, packaged to tarballs and placed into an artifact repository through a CI pipeline. This process is used for the internal configuration of our clusters (pipelines, indexes, inputs, authentication, access control) as well as user applications (e.g. dashboards and alerts).

Several of these applications were built with the help of Splunk PS, who generally recommended to put a small set of config files   into apps with a meaningful names to ease the life of Splunk support. So our base config common to all clusters is already composed of a dozen of small apps / tarballs.

We are also trying to avoid duplicating configuration. So the same "base" apps are loaded to many cluster, and only the necessary settings are managed through cluster specific apps. Precedence and "composition" are used to achieve this. For instance our authentication / access control can be composed of 4 apps:
* The base app defining all LDAP settings and access common to many clusters, deployed to both searchhead and indexers
* Optionally an app overriding the URL of LDAP server to connect through a LDAP proxy/cache in a specific cloud region (SH + IDX)
* An cluster specific app containing the LDAP bind credentials encrypted with the encryption key of the cluster (SH + IDX)
* An app defining the access control specific to a cluster, generally only deployed to searchheads

This type of solution is also used to define the volumes and default indexes paths and settings in a reusable app, separately from the specific indexes definition for a given cluster.

## Our understanding of the solution

* The apps are defined as a list in the ansible configuration
* The inventory "randomizes" the order of this list [by placing its content into a set in the inventory script](https://github.com/splunk/splunk-ansible/blob/e9b63e071aba2bdd6f38b2270e1389a8a136b66c/inventory/environ.py#L343)
[...]

PR #555, opened 2020-10-05
Changing internal environ parsing data structure from set to array

nwang92:
Partially addressing #545, particularly around app installation order. Although it shouldn't matter, it seems like the installation order can impact other apps. I would hazard a guess as to say this is some problem with Splunk itself - adding an app with certain auth/proxy information might inherently cause subsequent API calls to take longer. Ideally this doesn't happen, but not sure if there's some setting in the app itself that controls this or a problem in Splunk. 

In any case, preserving order does seem to mitigate it, so that's as much as we can do here.

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -340,19 +340,22 @@ def getSplunkApps(vars_scope):
     """
     Determine the set of Splunk apps to install as union of defaults.yml and environment variables
     """
-    appSet = set()
+    appList = []
     if not "apps_location" in vars_scope["splunk"]:
         vars_scope["splunk"]["apps_location"] = []
     # From default.yml
     elif type(vars_scope["splunk"]["apps_location"]) == str:
-        appSet.update(vars_scope["splunk"]["apps_location"].split(","))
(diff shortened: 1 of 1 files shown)


jrigassio-splunk (approved):

inventory/environ.py
```diff
@@ -340,19 +340,22 @@ def getSplunkApps(vars_scope):
     """
     Determine the set of Splunk apps to install as union of defaults.yml and environment variables
     """
-    appSet = set()
+    appList = []
```
bb03, line 343: Why an array?  I see you're removing duplicates later in this method.  Is it to preserve the original order?

bb03 (approved):

inventory/environ.py
nwang92: Yeah, per the issue, it seems like certain apps can impede REST API functionality when installed in a randomized order. I didn't look into the specifics of it - my guess is this might be things that intercept REST handlers or add proxying or change auth backends? 

In any case, I don't why we can't expose an ordered installation if that's what folks want.

PR #566, opened 2020-10-26
Fixed the when order in the hec_receiver task. If no token is defined…

schneewe:
… and hec should be disabled an error would occur, otherwise.
Now the tasks checks first the same condition like in the "Get existing HEC token" step.


nwang92:
Could you explain a little further the error you're hitting? This change seems to be causing some molecule test failures, so just want to see if the error is intentional + maybe we should add a new scenario to capture this change.

schneewe:
@nwang92 My I wanted to create a Splunk Server (in my case a AIO instance) but without HEC. Tried first without the complete setting, but without success. Then I tested with this settings:
```
splunk:
  hec:
      enable: false
```

If I now run the playbook, it doesn't activate the hec (as intended; have a. look at task set_as_hec_receiver.yml in the splunk_common role). 
The first step run and it creates and disabled hec input, the second step got skipped because of the when clause (I didn't want to create a HEC input, so I never configured a token in the yml file): 
`('hec' in splunk and 'token' in splunk.hec) or ('hec_token' in splunk)`

But the second step is a prerequisite of step 3's when clause, because the variable "hec_token" of step 2 need to be evaluated first -> this is the point I changed in my case/tests, because if I change the order of the when statements this step 3 got skipped too, due to I didn't configured a token value in the config yml. Afterwards the playbook run successful for me:
```
when:
    - hec_token.status == 200
    - ('hec' in splunk and 'token' in splunk.hec and hec_token.json['entry'][0]['content']['token'] != splunk.hec.token) or ('hec_token' in splunk and hec_token.json['entry'][0]['content']['token'] != splunk.hec_token)
```

bb03 (dismissed):

nwang92:
@schneewe got it, that makes sense! It seems like the test is failing because the existing order of
```
- hec_token.status == 200
- ('hec' in splunk and 'token' in splunk.hec and hec_token.json['entry'][0]['content']['token'] != splunk.hec.token) or ('hec_token' in splunk and hec_token.json['entry'][0]['content']['token'] != splunk.hec_token)
```
works to ensure that the JSON response is well-formed and has the necessary keys. When the order is reversed, a "bad" response from the GET may not guarantee the existence of the structure.

Maybe to encompass your use case, we can do the following?
```
- hec_token is defined and hec_token.status == 200
- ('hec' in splunk and 'token' in splunk.hec and hec_token.json['entry'][0]['content']['token'] != splunk.hec.token) or ('hec_token' in splunk and hec_token.json['entry'][0]['content']['token'] != splunk.hec_token)
```

That should also handle the skipped tasks that you're working around here.

schneewe:
@nwang92 ahh, yes that makes sense too.

Yeah completely fine. The fix for me was to change the order, but I didn't tested it with a set token. 
So to check the existence of the variable itself should solve both cases.

nwang92 (approved):

nwang92:
I thought I merged this a while ago but I didn't 😭 doing so now - thanks for the contribution!

schneewe:
> I thought I merged this a while ago but I didn't 😭 doing so now - thanks for the contribution!

No problem sir ;)

PR #568, opened 2020-10-29
Splunk Monitor update distsearch.conf

kashok-splunk:
- Removing stale distributed peers info from distsearch.conf
- Updating dmc groups after they have been created


kashok-splunk:
Will reopen

PR #573, opened 2020-11-13
Adding a small pause as it takes time for the distributed peer list to get updated

kashok-splunk:
Adding a small pause as it takes time for the distributed peer list to get updated


nwang92:
Static waits like this may not work well in distributed environments. There's no guarantee that Splunk will finish what you want it to do within the 3s wait here, especially if there are other processes/containers running on the same host. I'll conditionally approve, but just an FYI.

nwang92 (approved):

Issue #607, opened 2021-03-02
Support for Podman Rootless containers broke


PrymalInstynct:
## Issue
Between hub.docker.com container versions
- splunk/splunk:8.0.7 and splunk/splunk 8.0.8
OR
- splunk/splunk:8.1.1 and splunk/splunk 8.1.2

The ability to run the container via rootless podman execution stopped working.

It appears that the failure is being caused by the merge of `Feature/declarative password` #517
### Expected Outcome
Splunk container will start running as a rootless user

#### Ansible playbook
```
- hosts: localhost
  become: yes
  tasks:

  - name: "Create Rootless Container Users"
    user:
      name: "{{ item.name }}"
      comment: "{{ item.comment }}"
      uid: "{{ item.uid }}"
    with_items:
      - { name: 'splunkers', comment: 'Splunk Server', uid: '5008' }
  
  - name: "COPY | Setup .bashrc for Rootless Container Users"
    copy:
      src: '.bashrc'
      dest: "/home/{{ item }}"
    with_items:
      - splunkers
  
  - name: "Create ./ansible/tmp diretory for Splunk User"
    file:
      path: /home/splunkers/.ansible/tmp
      recurse: yes
      owner: splunkers
      group: splunkers
      mode: '0770'
  
  - name: "Check if Lingering is enabled for Splunk User"
    stat:
      path: "/var/lib/systemd/linger/splunk"
    register: splunk_lingering
    changed_when: no
  
  - name: "Enabled Lingering for Splunk User"
    command: loginctl enable-linger splunk
    when: not splunk_lingering.stat.exists
  
  - name: "Pull Image for Splunk"
    become: yes
    become_user: splunkers
    podman_image:
      name: splunk/splunk
      tag: '8.0.8'
      state: present
  
  - name: "Add Firewalld Rule for Splunk Container"
    firewalld:
      zone: public
      port: "{{ item }}"
      permanent: yes
      immediate: yes
      state: enabled
    with_items:
      - 8000/tcp
      - 8089/tcp
      - 9997/tcp
  
  - name: "Create Volume for Splunk Container"
    become: yes
    become_user: splunkers
    podman_volume:
      name: "{{ item }}"
      state: present
    with_items:
      - splunk_etc
      - splunk_var
  
  - name: "Create & Start Splunk Container"
    become: yes
    become_user: splunkers
    podman_container:
      name: splunk
      image: splunk/splunk:8.0.8
      state: started
      publish:
[...]

PR #613, opened 2021-03-16
Adding retry around restart check, timeouts around downloads, and use…

nwang92:
* Adding retry around restart check
* Adding timeouts around downloads
* Adding user ACL on user-seed.conf

I'm hoping the latter addresses https://github.com/splunk/splunk-ansible/issues/607

### DIFF
--- a/roles/splunk_common/tasks/check_for_required_restarts.yml
+++ b/roles/splunk_common/tasks/check_for_required_restarts.yml
@@ -11,6 +11,9 @@
     use_proxy: no
   register: restart_required
   changed_when: restart_required.status == 200
+  until: restart_required is succeeded
+  retries: 5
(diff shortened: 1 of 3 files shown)


jrigassio-splunk (approved):

roles/splunk_common/tasks/check_for_required_restarts.yml
```diff
@@ -11,6 +11,9 @@
     use_proxy: no
   register: restart_required
   changed_when: restart_required.status == 200
+  until: restart_required is succeeded
+  retries: 5
```
bb03, line 15: I'm a little worried here about the emergent behavior of this retry loop.  Like if there really is an underlying problem, it would be nice to fail faster.  Is it possible to enter this retry loop only on specific error conditions?

PR #628, opened 2021-06-07
Add apps_location_local for local apps on CM/Deployer

jryb:
### Problem

When a CM or a SH Deployer is deployed, there is no way to install a Splunk App locally on the CM/Deployer instance itself.  Currently all apps specified in `apps_location` are pushed to the cluster and disabled locally (see https://github.com/splunk/splunk-ansible/issues/545 for details on current functionality).

### Solution

Splunk ansible does have the functionality to install a Splunk app on a local Splunk instance, we just need to treat the CM/Deployer instance as a standalone.  Add a new variable, `apps_location_local`, that can be used in the CM/Deployer context to install apps locally only. This would leave the functionality of `apps_location` unchanged and add the ability to install local apps only if `apps_location_local` is configured.  On any other role besides CM or Deployer `apps_location_local` would do nothing.

### Tests

Mount 3 applications and test out the combinations of `apps_location` and `apps_location_local` with CM, SHC, and Standalone roles.

#### On a CM

default.yml config
```
  defaults: |-
    splunk:
      apps_location:
        - "/mnt/apps/application1.tgz"
        - "/mnt/apps/application2.tgz"
      apps_location_local:
        - "/mnt/apps/application3.tgz"
```
Verify application3 is enabled locally and not in bundle while application1 and application2 are in the bundle and disabled locally.
```
jryb:AppFramework$ kex-sc splunk-cm1-cluster-master-0  splunk
[splunk@splunk-cm1-cluster-master-0 splunk]$ bin/splunk display app
alert_logevent                 CONFIGURED          ENABLED             INVISIBLE           
alert_webhook                  CONFIGURED          ENABLED             INVISIBLE           
application1                   UNCONFIGURED        DISABLED            VISIBLE             
application2                   UNCONFIGURED        DISABLED            VISIBLE             
application3                   UNCONFIGURED        ENABLED             VISIBLE             
...          
SplunkLightForwarder           UNCONFIGURED        DISABLED            INVISIBLE           
[splunk@splunk-cm1-cluster-master-0 splunk]$ ls etc/master-apps/
_cluster  application1	application2
```

#### On a Deployer

default.yml config
```
  defaults: |-
    splunk:
      apps_location:
        - "/mnt/apps/application1.tgz"
        - "/mnt/apps/application2.tgz"
      apps_location_local:
        - "/mnt/apps/application3.tgz"
```
Verify application3 is enabled locally and not in bundle while application1 and application2 are in the bundle and disabled locally.
```
[splunk@splunk-shc1-deployer-0 splunk]$ bin/splunk display app
alert_logevent                 CONFIGURED          ENABLED             INVISIBLE           
alert_webhook                  CONFIGURED          ENABLED             INVISIBLE           
application1                   UNCONFIGURED        DISABLED            VISIBLE             
application2                   UNCONFIGURED        DISABLED            VISIBLE             
application3                   UNCONFIGURED        ENABLED             VISIBLE             
appsbrowser                    CONFIGURED          ENABLED             INVISIBLE           
...          
SplunkLightForwarder           UNCONFIGURED        DISABLED            INVISIBLE           
[splunk@splunk-shc1-deployer-0 splunk]$ ls etc/shcluster/apps/
README	application1  application2
```
Verify the SH bundle applied application1 and application2 on a SHC member
```
[splunk@splunk-shc1-search-head-0 splunk]$  bin/splunk display app
alert_logevent                 CONFIGURED          ENABLED             INVISIBLE           
alert_webhook                  CONFIGURED          ENABLED             INVISIBLE           
application1                   UNCONFIGURED        ENABLED             VISIBLE             
application2                   UNCONFIGURED        ENABLED             VISIBLE             
...        
SplunkLightForwarder           UNCONFIGURED        DISABLED            INVISIBLE           
[splunk@splunk-shc1-search-head-0 splunk]$ ls etc/apps/application*
etc/apps/application1:
bin  default  metadata

etc/apps/application2:
bin  default  metadata
[splunk@splunk-shc1-search-head-0 splunk]$
```

#### On a Standalone
default.yml config
```
  defaults: |-
    splunk:
      apps_location:
        - "/mnt/apps/application1.tgz"
        - "/mnt/apps/application2.tgz"
      apps_location_local:
        - "/mnt/apps/application3.tgz"
```
Verify that `apps_location_local` does nothing and only application1 and application2 are installed and enabled locally.
```
[splunk@splunk-sa-standalone-0 splunk]$ bin/splunk display app 
alert_logevent                 CONFIGURED          ENABLED             INVISIBLE           
alert_webhook                  CONFIGURED          ENABLED             INVISIBLE           
application1                   UNCONFIGURED        ENABLED             VISIBLE             
application2                   UNCONFIGURED        ENABLED             VISIBLE             
appsbrowser                    CONFIGURED          ENABLED             INVISIBLE           
...          
SplunkLightForwarder           UNCONFIGURED        DISABLED            INVISIBLE           
[splunk@splunk-sa-standalone-0 splunk]$ ls etc/apps/application*
etc/apps/application1:
bin  default  local  metadata

etc/apps/application2:
bin  default  local  metadata
```


inventory/environ.py
```diff
@@ -377,6 +377,24 @@ def getSplunkApps(vars_scope):
                 appList.append(app)
     vars_scope["splunk"]["apps_location"] = appList
 
+    # Check if theres a local apps list for CM/Deployer roles
```
nwang92, line 380: Maybe this section should be a separate function? That way we can add [test coverage](https://github.com/splunk/splunk-ansible/blob/develop/tests/small/test_environ.py) for it?

roles/splunk_common/tasks/provision_apps_local.yml
```diff
@@ -0,0 +1,23 @@
+---
```
nwang92, line 1: Can we reused the existing https://github.com/splunk/splunk-ansible/blob/develop/roles/splunk_common/tasks/provision_apps.yml, but modify it so that it accepts a variable as an input? Then we can pass in either `splunk.apps_location_local` or `splunk.apps_location` and reduce playbook duplication.

roles/splunk_deployer/tasks/main.yml
```diff
@@ -67,4 +67,8 @@
 - include_tasks: ../../../roles/splunk_common/tasks/disable_installed_apps.yml
   when: "'apps_location' in splunk and splunk.apps_location"
 
+# Now install local apps in apps_location_local exists
+- include_tasks: ../../../roles/splunk_common/tasks/provision_apps_local.yml
```
nwang92, line 71: Seeing that the disable_installed_apps.yml is called before this, does this guarantee that a local app is still enabled? Let's say the deployer happens to restart which will trigger the playbooks to run again. I'm not sure if we're preserving idempotency and as a result, the local app will end up disabled after the reboot.

inventory/environ.py
jryb: Yea, I'll move this to a separate function.

roles/splunk_common/tasks/provision_apps_local.yml
jryb: The local version of this basically the same.  I'll adjust the calls to the provision_apps.yml to handle either list.

roles/splunk_deployer/tasks/main.yml
jryb: I'll try this out with a deployer reset.  It is difficult to preserve the existing functionality of `apps_location` and add a local app install due to the fact that to deploy apps to the Cluster/SH bundle they are installed locally then copied to the bundle and disabled.  So this can lead to a case where, for example, let's say we have app1 and app2 for the SHC and app3 local.  What will happen is:

1. app1 and app2 are installed locally
2. locally installed apps are copied to the sh_bundle (including app1 and app2)
3. app1 and app2 are disabled
4. app3 is installed locally.

This works fine the first time around.  If we then run the playbook again what happens is:
1. app1 and app2 are installed locally
2.  locally installed apps are copied to the sh_bundle (including app1, app2, and app3)
3. app1, app2, and app3 are disabled locally
4. app3 is installed/re-enabled locally

Now app3 has migrated to the cluster.  This will actually happen with any locally installed/enabled app, installed through ansible or manually, when the playbook is run.  With the current method of installing all apps local and copying them to the bundle, I'm not sure there is a way around this.  We would have to change the way apps are unpacked/installed for the bundle which would likely effect the way premium apps such as ITSI and ES are installed and be a much larger diff.

nwang92: Yeah, that's the concern I had as well. The 2nd scenario is what I expect you'll hit with the operator/kubernetes, given the long-lived-ness of Splunk deployments and the possibility of pod restarts that can occur during the lifecycle of a each container. 

I heard there's some app management ERD/RFC around the operator, I'm not sure if this work is part of that initiative or if this is just a temporary workaround while those details are being refined. But I'm sure you're going to subject to the limitations of case 2 with this change.

nwang92 (approved):

roles/splunk_deployer/tasks/main.yml
jryb: Thanks Nelson!  Yes this `scope` (i.e. cluster or locally installed apps) will be used as part of the operators app installation framework.  For that work this fix would be used to install for the app installation when the Splunk containers are first brought up.  For "on the fly" update/install of apps, this would be an issue.

bb03 (approved):

PR #637, opened 2021-07-10
feat: Avoid use of chown to be more friendly to fixed uid/gid fs mounts

rfaircloth-splunk:
When using EFS in aws as the mount for etc/var the use of explicity group/user by ansible will fail as follows

```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "gid": 1037,
    "group": "1037",
    "mode": "0644",
    "owner": "1037",
    "path": "/opt/splunk/etc/.ui_login",
    "size": 0,
    "state": "file",
    "uid": 1037
}

MSG:

chown failed: [Errno 1] Operation not permitted: b'/opt/splunk/etc/.ui_login'
```

This change uses become user exclusively (safe). This also reduces behaviors that can trigger security tools for no essential reason.


roles/splunk_cluster_master/tasks/generate_ess_bundle.yml
```diff
@@ -37,8 +35,6 @@
   unarchive:
     src: "{{ ta_location.stdout_lines[0] }}"
     dest: "{{ splunk.app_paths.idxc }}"
```
nwang92, line 37: Will this also need the become/become_user since it's extracting files directly into some path within SPLUNK_HOME?

roles/splunk_common/tasks/add_forward_server.yml
```diff
@@ -14,20 +14,12 @@
     - "'forwarded-server already present' not in forward_status.stderr"
   no_log: "{{ hide_password }}"
 
-- name: "Adjust outputs.conf permissions"
```
nwang92, line 17: It looks like this was needed/added as part of our internal WIndows usage: https://github.com/splunk/splunk-ansible/pull/489

roles/splunk_cluster_master/tasks/generate_ess_bundle.yml
rfaircloth-splunk: yes I was container focused where we don't support ES/ITSI yet made adjustment

roles/splunk_common/tasks/add_forward_server.yml
rfaircloth-splunk: yes I was container focused where we don't support ES/ITSI yet made adjustment

roles/splunk_cluster_master/tasks/generate_ess_bundle.yml
jmeixensperger: Was there a workaround added for this fix? I'll kick off a windows ec2 test to verify as well.

roles/splunk_common/tasks/add_forward_server.yml
jmeixensperger: Was there a workaround added for this fix? I'll kick off a windows ec2 test to verify as well.

PR #687, opened 2022-05-19
if exists, delete passwd file from splunk home;

adityapinglesf:


roles/splunk_common/tasks/enable_admin_auth.yml
```diff
@@ -28,6 +28,11 @@
     become_user: "{{ splunk.user }}"
     no_log: "{{ hide_password }}"
 
+  - name: "Delete etc/passwd if exists (Windows)"
```
hendolim, line 31: This will actually apply to any OS. There's not condition set to limit it to Windows.
Might want to change the name to not mislead people.

adityapinglesf: updating the ansible code to match up for windows only

jmeixensperger (approved):

sarahotis (approved):

PR #704, opened 2022-12-20
CSPL-2200: fix ansible code to not update web.conf when set to "false"

gaurav-splunk:
Fix ansible code to not update web.conf in etc/system/local/web.conf when not explicitly set to true.

### DIFF
--- a/roles/splunk_common/tasks/main.yml
+++ b/roles/splunk_common/tasks/main.yml
@@ -93,6 +93,7 @@
 - include_tasks: enable_splunkweb_ssl.yml
   when:
     - "'http_enableSSL' in splunk and splunk.http_enableSSL is not none"
+    - splunk.http_enableSSL | bool
 
 - include_tasks: enable_splunkd_ssl.yml
   when: "'ssl' in splunk and splunk.ssl"


roles/splunk_common/tasks/main.yml
```diff
@@ -130,4 +131,4 @@
 - include_tasks: add_splunk_license.yml
 
 - include_tasks: disable_popups.yml
-  when: "'disable_popups' in splunk and splunk.disable_popups | bool"
\ No newline at end of file
+  when: "'disable_popups' in splunk and splunk.disable_popups | bool"
```
akondur, line 134: Was this change intentional?

akondur (approved):
Looks good

kumarajeet (approved):
Approving assuming @akondur comment is validated.

roles/splunk_common/tasks/main.yml
gaurav-splunk: nope I don't think so I made any change in that.

akondur: Looks like a few extra spaces might have been removed on line 135 and that is fine. Closing the comment thread now.

alishamayor (approved):

PR #724, opened 2023-04-25
CSPL-2353 - Bias lang update

akondur:
Change `master-uri` to `manager-uri` & `slave` to `peer`


adityapinglesf (dismissed):

sgontla:
Just for my understanding from the Operator point of view, do we need to tie any Operator changes to this? What happens if the customer just updates the Splunk docker image with this changes? Also, If we have any Operator specific changes(?), what happens if the customer just updates the Operator alone?

akondur:
> Just for my understanding from the Operator point of view, do we need to tie any Operator changes to this? What happens if the customer just updates the Splunk docker image with this changes? Also, If we have any Operator specific changes(?), what happens if the customer just updates the Operator alone?

@sgontla Good point. We are changing only the cli and conf file changes syntax here from master to manager. As far as operator is concerned we should be able to continue using the same config and ansible underneath will use manager vs master to configure splunkd. 

> Just for my understanding from the Operator point of view, do we need to tie any Operator changes to this? What happens if the customer just updates the Splunk docker image with this changes? Also, If we have any Operator specific changes(?), what happens if the customer just updates the Operator alone?

PR #727, opened 2023-05-17
Update wait_for_splunk_process.yml

bendermar:
We need an increase of this timeout because with large searchhead cluster with lot of users and apps this results into a loop during upgrade or scaling


adityapinglesf (approved):
can we add a TIMEOUT_VALUE variable here?

michaellee-splunk (approved):

PR #735, opened 2023-06-12
update from deprecated callback_whitelist into callbacks_enabled

bilalbayasut:
when using this playbooks , I encountered
```
[DEPRECATION WARNING]: [defaults]callback_whitelist option, normalizing names 
to new standard, use callbacks_enabled instead. This feature will be removed 
from ansible-core in version 2.15. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.
```
so this PR basically update the `ansible.cfg` configs based on that feedback, so the warning message is suppressed.


bilalbayasut:
@adityapinglesf hi, any review regarding this ? thanks

abdulk-splunk (approved):
Ansible Doc Reference: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#callbacks-enabled

PR #765, opened 2023-12-05
INFRA-40617: Custom Auth Module for HTTP Request

abdulk-splunk:


roles/splunk_common/tasks/install_apps.yml
```diff
@@ -31,13 +31,10 @@
...
   - name: Download remote app
-    get_url:
+    http_auth:
       url: "{{ app_url }}"
       dest: /tmp/
-      mode: 0777
-      timeout: 120
-      validate_certs: no
-      force: yes
+      token: "{{ lookup('env', 'ARTIFACTORY_AUTHORIZATION') }}"
```
hendolim, line 37: We should really use `get_url` here rather than having to maintain a custom module. If the only reasoning is due to header redirection, we should work on getting Orca to update ansible so we can use `undirected_headers`: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-unredirected_headers

Taking note to track it as a tech debt

hendolim (approved):

ruomeiy-splunk (approved):

PR #777, opened 2024-01-23
Changes to support Unix Domain Socket

adityapinglesf:
Changes to support Unix Domain Socket. All API calls will check if UDS is enabled. If enabled, it will perform requests via unix socket. If not enabled, it will perform requests over regular SVC_PORT.

### DIFF
--- a/library/splunk_api.py
+++ b/library/splunk_api.py
@@ -1,0 +1,82 @@
+#!/usr/bin/python
+
+from ansible.module_utils.basic import AnsibleModule
+import os
+import requests
+import requests_unixsocket
+import json
+
+UDS_SOCKET_PATH = "../var/run/splunk/cli.socket"
+
+def supports_uds():
+    return os.path.exists(UDS_SOCKET_PATH)
+
+def api_call_uds(method, endpoint, username, password, payload=None, headers=None, verify=False, status_code=None, timeout=None):
+    session = requests_unixsocket.Session()
+    url = f"http+unix://{UDS_SOCKET_PATH}{endpoint}"
+    if headers is None:
+        headers = {}
+    headers['Content-Type'] = 'application/json'
+    auth = (username, password)
+
+    response = session.request(method, url, headers=headers, auth=auth, data=json.dumps(payload), verify=verify, timeout=timeout)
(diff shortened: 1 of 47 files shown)


library/splunk_api_test.py
```diff
@@ -0,0 +1,83 @@
...
+    endpoint = module.params['url']
+    username = module.params['username']
+    password = module.params['password']
+    payload = module.params.get('payload', None)
+    headers = module.params.get('headers', None)
+    verify = module.params.get('verify', False)
+    status_code = module.params.get('status_code', None)
+    timeout = module.params.get('timeout', None)
+
+    if supports_uds():
```
adityapinglesf, line 71: default: 8089 port, even if UDS is enabled. need a flag to switch between UDS/HTTP

list of actions:

- build
- MUST deploy with UDS and non-UDS env
- Splunk/UF works as expected

library/splunk_api.py
adityapinglesf, line 1: line 9 and 10 needs to be in the constants file

adityapinglesf: UDS socket path needs to be relative -- not hardcoded for  opt as {splunk.home}

adityapinglesf: can it be configured and do we want splunk users to configure this??

roles/splunk_common/tasks/set_mgmt_mode.yml
```diff
@@ -0,0 +1,20 @@
...
+    owner: "{{ splunk.user }}"
+    group: "{{ splunk.group }}"
+  when: splunk.splunk_http_enabled|bool == false
+
+- name: Set Mgmt Mode TCP
+  ini_file:
+    dest: "{{ splunk.home }}/etc/system/local/server.conf"
+    option: "mgmtMode"
+    section: "httpServer"
+    value: "auto"
```
hendolim, line 17: This value is the same as line 6?

library/splunk_api.py
adityapinglesf: flag should be enabled by default (as it's latest feature)
end user should opt out irrespective of UDS capability of the system.

make it specific - uds or tcp

PR #791, opened 2024-03-08
Add ansible collection support

tod-uma:
## Why

This makes it possible to consume `splunk-ansible` as an Ansible collection — installed via `ansible-galaxy collection install` from a git-sourced `requirements.yml` — instead of vendoring or symlinking the repo into a project's roles path. That's useful for anyone integrating splunk-ansible's roles into their own playbooks (for example, installing the universal forwarder as one step of a broader Ansible-managed bootstrap process) without maintaining a private fork just to get collection semantics.

Ansible collections need two things this repo doesn't have: a `galaxy.yml` manifest, and modules living under `plugins/modules/` rather than `library/` (`library/` is the pre-collection, role-local module path; a collection's own modules resolve from `plugins/modules/` — see [Developing collections](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections.html)).

## What

- **`galaxy.yml`** (new) — collection manifest. `description`, `license` (matches the repo's actual Apache-2.0 license), `tags`, and `documentation`/`homepage`/`issues` links to the existing docs site, repo, and issue tracker. `version` starts at `1.0.0`: a collection's version should track its own release history, not whatever Splunk build happens to be current, so it isn't tied to a product version number.
- **`library/` → `plugins/modules/`** — moves `splunk_api.py` and `es_ta_for_indexers.py` to the path a collection resolves its modules from. Done as renames (`git mv`), so history is preserved on both files.

## Verified

`ansible-galaxy collection build` succeeds and produces a valid `splunk-splunk_ansible-1.0.0.tar.gz`.

Once installed as a collection:

```yaml
# collections/requirements.yml
collections:
  - source: https://github.com/splunk/splunk-ansible.git
    version: "10.4.3"   # pin a release tag, not a moving branch
    type: git
```

```yaml
# playbook.yml
- hosts: splunk_universal_forwarder
  become: true
  collections:
    - splunk.splunk_ansible
  vars:
    splunk: "{{ splunk_universal_forwarder }}"
  tasks:
    - name: install the forwarder
      ansible.builtin.include_role:
        name: splunk_universal_forwarder
```

## Status

Opened in 2024 and stale since; rebased now onto current `develop` with no conflicts — this range hasn't touched `library/` or `galaxy.yml`, so both structural changes still apply cleanly. We've been running this on a fork since, alongside several other fixes that have since merged cleanly here (#879, #883, #893, #894), so happy to adjust anything in `galaxy.yml` — tags especially — to whatever fits the project's own preference.


tod-uma:
Rebased onto current `develop` (no conflicts — this range hasn't touched `library/` or `galaxy.yml`), filled in `galaxy.yml`'s description/license/tags/links, and reset the version to `1.0.0` so it tracks the collection's own history rather than a Splunk build number. Description above has the details.

We've been running this on a fork for a while now, alongside several other small fixes that have since merged cleanly here — #879, #883, #893, #894 — so figured it was worth bringing this one back into view too. Happy to adjust anything, `galaxy.yml`'s tags especially, to match project preference.

PR #795, opened 2024-03-12
Optimize deployment setup time

jonathan-vega-splunk:
Removed multiple restarts that were slowing down deployment setup as well as other tasks that are no longer recommended per Splunk Docs.

### DIFF
--- a/roles/splunk_common/tasks/main.yml
+++ b/roles/splunk_common/tasks/main.yml
@@ -119,13 +119,9 @@
 - include_tasks: enable_dsp.yml
   when: "'dsp' in splunk and 'enable' in splunk.dsp and splunk.dsp.enable"
 
-- include_tasks: start_splunk.yml
-
-- include_tasks: set_certificate_prefix.yml
-
-- include_tasks: clean_user_seed.yml
-
 - include_tasks: add_splunk_license.yml
 
 - include_tasks: disable_popups.yml
   when: "'disable_popups' in splunk and splunk.disable_popups | bool"
+
+- include_tasks: start_splunk.yml
--- a/roles/splunk_search_head/tasks/search_head_clustering.yml
+++ b/roles/splunk_search_head/tasks/search_head_clustering.yml
@@ -71,16 +71,3 @@
   retries: "{{ shc_sync_retry_num }}"
   delay: "{{ retry_delay }}"
   no_log: "{{ hide_password }}"
-
-- name: Destructive sync search head
-  command: "{{ splunk.exec }} resync shcluster-replicated-config -auth {{ splunk.admin_user }}:{{ splunk.password }}"
-  become: yes
-  become_user: "{{ splunk.user }}"
-  when: not splunk_search_head_captain | bool
-  register: task_result
-  changed_when: task_result.rc == 0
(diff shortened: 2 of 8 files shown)


roles/splunk_search_head/tasks/search_head_clustering.yml
```diff
@@ -75,16 +75,3 @@
   retries: "{{ shc_sync_retry_num }}"
   delay: "{{ retry_delay }}"
   no_log: "{{ hide_password }}"
-
-- name: Destructive sync search head
```
adityapinglesf, line 79: should we instead make it optional? would it cause any issues while performing upgrades?

roles/splunk_common/tasks/main.yml
```diff
@@ -122,14 +122,9 @@
 - include_tasks: enable_dsp.yml
   when: "'dsp' in splunk and 'enable' in splunk.dsp and splunk.dsp.enable"
 
-- include_tasks: start_splunk.yml
-
-- include_tasks: set_certificate_prefix.yml
-  when: splunk.splunk_http_enabled|bool == true
-
-- include_tasks: clean_user_seed.yml
```
adityapinglesf, line 130: do we need to remove this playbook from the repository? it's not used anywhere else.

roles/splunk_common/tasks/main.yml
```diff
@@ -122,14 +122,9 @@
 - include_tasks: enable_dsp.yml
   when: "'dsp' in splunk and 'enable' in splunk.dsp and splunk.dsp.enable"
 
-- include_tasks: start_splunk.yml
-
-- include_tasks: set_certificate_prefix.yml
```
adityapinglesf, line 127: also this playbook?

roles/splunk_common/tasks/main.yml
hendolim: I don'r recall exactly, but I think the user seed is used to determine if this is the first time Splunk is bootstrapped or not. I remember running into a problem when doing upgrade and when using local dev build because of it

roles/splunk_search_head/tasks/search_head_clustering.yml
jonathan-vega-splunk: This is one of those tasks that took almost 5 minutes to complete. According to this doc https://docs.splunk.com/Documentation/Splunk/9.2.0/DistSearch/HowconfrepoworksinSHC#Why_a_recovering_member_might_need_to_resync_manually we shouldn't need to perform a destructive sync unless the cluster is recovering a member, and then it's only necessary on that one instance. We always could leave this off by default and have some flag to run it if possible.

roles/splunk_common/tasks/main.yml
jonathan-vega-splunk: Yeah this one I don't have strong preferences about, I couldn't find anything in the docs about what happens if we keep this around.

roles/splunk_common/tasks/main.yml
jonathan-vega-splunk: This playbook, set_certificate_prefix I essentially incorporated into roles/splunk_common/tasks/enable_splunkd_ssl.yml. If we turn on ssl we might as well set the certificate prefix in the same file instead of having two different steps.

PR #802, opened 2024-03-20
use alternate register when ubuntu platform is used;

adityapinglesf:


roles/splunk_common/tasks/check_mgmt_mode_status.yml
```diff
@@ -7,14 +7,19 @@
 - name: Check if listening on SVC Port {{ splunk.svc_port }}
```
hendolim, line 7: I'd suggest to get rid of this task. Since this will add more dependency and we should keep splunk-ansible OS-agnostic

roles/splunk_common/tasks/check_mgmt_mode_status.yml
```diff
@@ -7,14 +7,19 @@
...
   debug:
     msg: "This UF instance is configured to use UDS socket for API communications."
-  when: client_socket_file.stat.exists and port_status.stdout.find("{{ splunk.svc_port }}") == -1
+  when: (client_socket_file.stat.exists and port_status.skipped is false and port_status.stdout.find("{{ splunk.svc_port }}") == -1) or (client_socket_file.stat.exists and port_status_ubuntu.skipped is false and port_status_ubuntu.stdout.find("{{ splunk.svc_port }}") == -1)
 
 - name: UF is configured to use TCP
   debug:
     msg: "This UF instance is configured to use TCP socket for API communications."
-  when: not client_socket_file.stat.exists and port_status.stdout.find("{{ splunk.svc_port }}") != -1
+  when: (not client_socket_file.stat.exists and port_status.skipped is false and port_status.stdout.find("{{ splunk.svc_port }}") != -1) or (not client_socket_file.stat.exists and port_status_ubuntu.skipped is false and port_status_ubuntu.stdout.find("{{ splunk.svc_port }}") != -1)
```
hendolim, line 25: We can probably just infer if UF is configured to use UDS or TCP by seeing if Splunk has produced the socket file. We just keep our business simple and don't actually go check if any HTTP port has been opened. We just assume that is the case because otherwise splunk-ansible wouldn't have been able to complete provisioning

roles/splunk_common/tasks/check_mgmt_mode_status.yml
adityapinglesf: done, removed netstat check

Issue #815, opened 2024-04-16
create_hec_token is not triggered is not added if "splunk_hec_token" does not already exist 


jackskj:
Background: 
i was updating splunk HEC ([splunk docker image](https://hub.docker.com/r/splunk/splunk/)) from 9.1.1 to 9.2.1.
in 9.1.1, SPLUNK_HEC_TOKEN envar is set and ansible is using it to create splunk_hec_token in create_hec_token step. 

what was expected: 
when updating to splunk 9.2.1, SPLUNK_HEC_TOKEN envar in create_hec_token to create the token just like in 9.1.1 

what actually happened:
in 9.2.1, create_hec_token is never triggered. 

Reason: 
i have noticed that ["Get existing HEC token" ](https://github.com/splunk/splunk-ansible/blob/fa8c30800276fae3c13e61d9261d482656e497db/roles/splunk_common/tasks/set_as_hec_receiver.yml#L35) step fetches "splunk_hec_token". However, this process fails as this token does not exist.
in 9.1.1, splunk_hec_token was triggered anyway, however, [this commit ](https://github.com/splunk/splunk-ansible/commit/fa8c30800276fae3c13e61d9261d482656e497db#diff-0d690ee20a93d110b301d314b36902ae0da2182f1166f022cd38ce4fb054b532) , a "hec_token is defined and hec_token.status == 200" [statement is added](https://github.com/splunk/splunk-ansible/blob/fa8c30800276fae3c13e61d9261d482656e497db/roles/splunk_common/tasks/set_as_hec_receiver.yml#L80) which prevents splunk_hec_token from being created, which to me is a "catch 22" 

to fix this issue, i am adding "http://splunk_hec_token" via [custom configs ](https://splunk.github.io/docker-splunk/ADVANCED.html#create-custom-configs)
```
splunk:
  conf:
    - key: inputs
      value:
        directory: /opt/splunk/etc/system/local
        content:
          "http://splunk_hec_token":
             token: dummytoken
```

with this added, "Get existing HEC token" successes, create_hec_token is triggered, and "dummytoken" is updated with correct token

aukevanleeuwen:
For our use case it would be reasonably difficult to get custom configuration going (over what we have now, just running it with environment variables). So if this could get fixed, that would be awesome.

PR #819, opened 2024-04-18
Fix windows bind/UDS regressions

jmeixensperger:
- Fix windows bind regression caused by: https://github.com/splunk/splunk-ansible/pull/737
- Add check for empty UDS cli.socket file, adjust `splunk_http_enabled` var
- Fix HEC token provisioning (https://github.com/splunk/splunk-ansible/issues/815)
- Add check/condition for installed version when setting `mgmtMode` (not available pre 9.1.0)


ruomeiy-splunk (approved):

jmeixensperger:
Local tests are passing with 9.0.9 and 9.3.0, ready for review

jmeixensperger:
@hendolim I verified that windows 9.3.0 has `mgmtHostPort = 127.0.0.1:8089` as the /default/web.conf value, so we'll need that separate play for windows / 127.0.0.1

hendolim (approved):

PR #836, opened 2024-05-14
checks added for hec configuration

BigCaber:
changes


jmeixensperger:
Can you please change the order so that the `is defined` check is performed first? If that var is not defined, the first check will fail to evaluate.

BigCaber:
reordered

BigCaber:
@jmeixensperger I think you missed my comment

jmeixensperger (approved):

PR #844, opened 2024-06-13
use separte set_fact for bearer token or basic auth download

tod-uma:
when trying to install a tgz from a no auth url, I ran into a bug where the task "Define a path for the downloaded package" would fail because `download_result` did not contain a key `dest`. This is because the `download_result` variable gets overwritten by the basic_auth task even though it gets skipped. This patch simply provides a separate "Define a path for the downloaded package" task for each download task.


tod-uma:
fixes #845

jmeixensperger (approved):

PR #855, opened 2024-07-18
exclude ds from disabled indexing

vanillaxx:
Indexing shouldn't be disabled for deployment_server since it leads to empty list of clients. Adjusted `when` condition so that it does not execute `Disable indexing on the current node` task if `splunk.role` is `deployment_server`.
image

### DIFF
--- a/roles/splunk_common/tasks/enable_forwarding.yml
+++ b/roles/splunk_common/tasks/enable_forwarding.yml
@@ -63,7 +63,9 @@
     owner: "{{ splunk.user }}"
     group: "{{ splunk.group }}"
   register: index_disabling
-  when: splunk.role != "splunk_license_master"
+  when:
+    - splunk.role != "splunk_license_master"
+    - splunk.role != "splunk_deployment_server"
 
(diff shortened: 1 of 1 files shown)


jmeixensperger (dismissed):

roles/splunk_common/tasks/enable_forwarding.yml
```diff
@@ -63,7 +63,9 @@
     owner: "{{ splunk.user }}"
     group: "{{ splunk.group }}"
   register: index_disabling
-  when: splunk.role != "splunk_license_master"
+  when:
+    - splunk.role != "splunk_license_master"
+    - splunk.role != "deployment_server"
```
jmeixensperger: I believe the role name should be "splunk_deployment_server"

vanillaxx: Applied changes

jmeixensperger (approved):

ruomeiy-splunk (approved):

PR #875, opened 2025-04-03
INFRA-43719: Add retry to reduce transient issues

ruomeiy-splunk:
Multisite deployment has a small chance of getting some race conditions when modifying directory ownership on site A, and site B is pushing some files to site A. Also adding retry to apply cluster bundle step, so it can handle the case when peers are restarting.

### DIFF
--- a/roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml
+++ b/roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml
@@ -14,36 +14,30 @@
   retries: "{{ retry_num }}"
   delay: "{{ retry_delay }}"
 
-# Fails if a cluster bundle did not already exist and yet we did not apply a new bundle.
-# The failed condition is basically eventHasError && bundleNotAlreadyExists.
-# The changed condition is basically !eventHasError && bundleNotAlreadyExists.
-#
 # Stderr/Stdout cases are:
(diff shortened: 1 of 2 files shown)


roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml
```diff
@@ -14,36 +14,30 @@
...
-      or ("bundle" not in splunk_cluster_bundle_result.stdout)
-      or (splunk_cluster_bundle_result.stdout is search("[Ee]rror")))
-    and ("already" not in splunk_cluster_bundle_result.stderr)
-  changed_when: >
+  until: >
     ((splunk_cluster_bundle_result.stdout is search("[Aa]pply"))
       and ("bundle" in splunk_cluster_bundle_result.stdout)
       and (splunk_cluster_bundle_result.stdout is not search("[Ee]rror")))
-    and ("already" not in splunk_cluster_bundle_result.stderr)
+    or ("already" in splunk_cluster_bundle_result.stderr)
```
jmeixensperger, line 40: Do you know when this condition is supposed to occur? Is it when the cluster bundle is already in sync?

jmeixensperger (approved):

hendolim (approved):

PR #881, opened 2025-07-30
CSPL-3866: Add option to avoid CM bundle push

hamzaaziz-splunk:
Change description: Option added to ansible to avoid bundle push that happens on cluster master. 

Issue ticket: https://splunk.atlassian.net/browse/CSPL-3866

### DIFF
--- a/roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml
+++ b/roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml
@@ -10,9 +10,7 @@
     status_code: [200]
   register: peer_list
   until: peer_list.json.entry | length == groups['splunk_indexer'] | length
-  when: 
-    - "'splunk_indexer' in groups"
-    - not (splunk.skip_cluster_bundle_push | default(false) | bool)
+  when: "'splunk_indexer' in groups"
   retries: "{{ retry_num }}"
   delay: "{{ retry_delay }}"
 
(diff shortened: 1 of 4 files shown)


roles/splunk_cluster_master/tasks/main.yml
```diff
@@ -53,5 +53,9 @@
     - splunk.multisite_search_factor_origin is defined
     - splunk.multisite_search_factor_total is defined
 
+- include_tasks: apply_cluster_bundle.yml
+  when: 
+    - not splunk.skip_cluster_bundle_push | default(false) | bool
```
jmeixensperger: Small nitpick: you don't need the `default(false)` part since your environ.py code already defaults the variable to false

roles/splunk_cluster_master/tasks/main.yml
```diff
@@ -53,5 +53,9 @@
     - splunk.multisite_search_factor_origin is defined
     - splunk.multisite_search_factor_total is defined
 
+- include_tasks: apply_cluster_bundle.yml
```
jmeixensperger: This task is already included on line 24 of this file. Please move the `when:` condition under that line and remove this block.

roles/splunk_cluster_master/tasks/apply_cluster_bundle.yml
```diff
@@ -10,7 +10,9 @@
     status_code: [200]
   register: peer_list
   until: peer_list.json.entry | length == groups['splunk_indexer'] | length
-  when: "'splunk_indexer' in groups"
+  when: 
+    - "'splunk_indexer' in groups"
+    - not (splunk.skip_cluster_bundle_push | default(false) | bool)
```
jmeixensperger: This condition will never be used properly because it is already checked before the file gets included (because of your `main.yml` change). In my opinion, you can revert all the changes in this file and leave it as is. The condition in `main.yml` should be good enough.

jmeixensperger:
Please add an entry in docs/ADVANCED.md for the new environment variable as well.

jmeixensperger (approved):

michaellee-splunk (approved):

PR #882, opened 2025-08-06
CPSL-3916: Add Ingestor Role

Igor-splunk:


michaellee-splunk (dismissed):

jmeixensperger:
You may want to name the role with the `splunk_` prefix so it's clear that this role is defined/supported by Splunk.

Otherwise, I just have a couple questions about the new role:
- I've seen mention of a new ingestor cluster (similar to search head/indexer clustering). Do we need to support that?
- Do we need to configure how to forward data along from the ingestor to other roles or services (Smart Bus / Smart Store)?

kasiakoziol:
> You may want to name the role with the `splunk_` prefix so it's clear that this role is defined/supported by Splunk.
> 
> Otherwise, I just have a couple questions about the new role:
> 
> * I've seen mention of a new ingestor cluster (similar to search head/indexer clustering). Do we need to support that?
> * Do we need to configure how to forward data along from the ingestor to other roles or services (Smart Bus / Smart Store)?

@jmeixensperger Let me push a fix to add "splunk_" prefix. 
-> Initially, the role is going to be the same as for the standalone instance. However, we want to copy it into a separate role to be able to customize it in the future based on our needs.
-> There is no need to configure that.

jmeixensperger (approved):

michaellee-splunk (approved):

PR #899, opened 2026-02-26
Add support for site-specific multisite SF and RF

rsivaprakas:
This PR adds support for site-specific multisite factors for both replication and search, while keeping current behaviour fully compatible.

Users can now provide optional site-specific terms like - `site1:1,site2:1` and the code will build the final values as - `origin:<origin>,<sites>,total:<total>`

So users can keep using existing `origin`/`total` settings and only add per-site terms when needed.

Changes:

- Added new optional inputs:
  - `SPLUNK_MULTISITE_REPLICATION_FACTOR_SITES`
  - `SPLUNK_MULTISITE_SEARCH_FACTOR_SITES`
  - YAML equivalents:
    - `multisite_replication_factor_sites`
    - `multisite_search_factor_sites`
- Moved final factor string construction into `inventory/environ.py` (single place of truth).
- Updated cluster manager setup to consume the computed final strings directly.
- Updated defaults and docs.
- Added tests for:
  - valid/invalid site-term parsing
  - final string building
  - legacy behaviour (origin/total only)
  - env and YAML site-specific inputs

Design options considered 

- High complexity: full free-form string input (`origin:...,site1:...,total:...`)
  - Flexible, but more duplication and easier to misconfigure.
- Medium complexity: map-based input in YAML
  - Structured, but more code and not ideal for current env-driven usage.
- Chosen (lowest complexity / clearest): keep existing `origin` + `total`, add optional `*_SITES` only
  - Least disruptive, easiest to understand, and backward compatible.

Backward compatibility

No breaking change.

If users only set existing origin/total variables, behaviour stays the same.

Validation done

- Ran targeted tests:
  - `python3 -m pytest tests/small/test_environ.py -k multisite -q`
  - Result: `12 passed`
- Lint checks on changed files: no linter errors.

### DIFF
--- a/docs/ADVANCED.md
+++ b/docs/ADVANCED.md
@@ -94,8 +94,10 @@ Splunk-Ansible ships with an inventory script in `inventory/environ.py`. The scr
 | SPLUNK_MULTISITE_MASTER_PORT | For multisite topologies, define the Splunk management port of the multisite cluster master | no | no | no |
 | SPLUNK_MULTISITE_REPLICATION_FACTOR_ORIGIN | For multisite topologies, define the origin replication factor | no | no | no |
 | SPLUNK_MULTISITE_REPLICATION_FACTOR_TOTAL | For multisite topologies, define the total replication factor | no | no | no |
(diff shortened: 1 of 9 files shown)


daniel-choi-158 (approved):
Thanks for this, adds an important layer of granular control

jmeixensperger (approved):

hendolim (approved):

PR #901, opened 2026-04-02
Fix SPLUNK_LISTEN_ON_IPV6="false" being treated as truthy

ruomeiy-splunk:
## Summary
- `SPLUNK_LISTEN_ON_IPV6="false"` passed via `--env` is a non-empty string, which is truthy in Python — causing `mgmtHostPort` in `web.conf` to be set to `[::1]:8089` (IPv6) even when the user intended to disable IPv6.
- Fixed by normalizing with `.lower() == "true"`, consistent with how other boolean env vars are handled in `environ.py` (e.g. `SPLUNK_SKIP_CLUSTER_BUNDLE_PUSH`).
- Bug introduced in commit `1c45c9d` (Oct 7, 2024).

## Root cause
In `inventory/environ.py`, `getIPv6()` stored the raw string from `os.environ.get()`:
```python
# Before (buggy): "false" is a non-empty string → truthy
vars_scope["splunk"]["listen_on_ipv6"] = os.environ.get("SPLUNK_LISTEN_ON_IPV6", False)

# After (fixed): explicit boolean parse
vars_scope["splunk"]["listen_on_ipv6"] = os.environ.get("SPLUNK_LISTEN_ON_IPV6", "").lower() == "true"
```

## Test plan
- [ ] Deploy with `--env SPLUNK_LISTEN_ON_IPV6="false"` on Splunk 10.2+ and verify `mgmtHostPort` is `0.0.0.0:8089`
- [ ] Deploy with `--env SPLUNK_LISTEN_ON_IPV6="true"` and verify `mgmtHostPort` is `[::1]:8089`
- [ ] Deploy without the env var and verify `mgmtHostPort` is `0.0.0.0:8089`

### DIFF
--- a/inventory/environ.py
+++ b/inventory/environ.py
@@ -739,7 +739,7 @@ def getIPv6(vars_scope):
     """
     Set specific environment variables to apply IPv6 configurations
     """
-    vars_scope["splunk"]["listen_on_ipv6"] = os.environ.get("SPLUNK_LISTEN_ON_IPV6", False)
+    vars_scope["splunk"]["listen_on_ipv6"] = os.environ.get("SPLUNK_LISTEN_ON_IPV6", "").lower() == "true"
 
 
 def getRandomString():


jmeixensperger:
I don't see why the original code is an issue. My understanding is:
1. The inventory script sets the `splunk.listen_on_ipv6` to a string value of "false"
2. In the `splunk_common` main.yml file, we use a `| bool` filter to evaluate the value:
```
- include_tasks: enable_ipv6.yml
  when: splunk.listen_on_ipv6 | bool
```

That filter/condition should evaluate the "false" string as False and skip the play. Is there something I'm missing/wrong about?

ruomeiy-splunk:
You're right that `enable_ipv6.yml` is correctly guarded by `| bool` in `main.yml`. However, there's a second code path that is **not** protected: `set_mgmt_port.yml`, which is called **unconditionally** from `main.yml` and uses a raw Jinja2 `{% if %}` without `| bool`:

```yaml
# roles/splunk_common/tasks/set_mgmt_port.yml:4
localhost_address: "{% if splunk.listen_on_ipv6 %}{{ '[::1]' }}{% else %}{{ '0.0.0.0' }}{% endif %}"
```

In Jinja2, `{% if "false" %}` evaluates to `True` because `"false"` is a non-empty string. This is what causes `mgmtHostPort` to be set to `[::1]:8089` even when the user passes `SPLUNK_LISTEN_ON_IPV6="false"`.

So the two code paths behave differently:
- `enable_ipv6.yml` — guarded by `splunk.listen_on_ipv6 | bool` ✅ correctly skipped
- `set_mgmt_port.yml` — raw `{% if splunk.listen_on_ipv6 %}` ❌ string `"false"` is truthy → sets `[::1]:8089`

We could alternatively fix `set_mgmt_port.yml:4` to use `{% if splunk.listen_on_ipv6 | bool %}`, but fixing it at the Python source ensures the value is always a proper boolean and prevents the same class of bug if more Jinja2 conditions on `listen_on_ipv6` are added in the future without a `| bool` guard.

jmeixensperger (approved):

hendolim (approved):

PR #905, opened 2026-05-28
CSPL-4006: Fix SSL replication port deleted on first indexer cluster join

gabrielm-splunk:
## Problem

Customers configuring SSL replication port via `splunk.conf.server.content` (e.g. `replication_port-ssl://9887`) find the stanza deleted after the indexer joins the cluster on first startup. The SSL config only persists after a **second pod restart**, requiring manual intervention. On some published `splunk/splunk` base images, the peer can also end up permanently unregistered from the ClusterManager even though the pod reports healthy (see Fix 3 below).

### Customer Configuration
```yaml
splunk:
  conf:
    server:
      content:
        'replication_port-ssl://9887':
          serverCert: /mnt/peers-splunk-cert/tls.crt
          sslVersions: tls1.2
          disabled: false
```

## Root Cause

Three issues stack on top of each other in the indexer cluster-join path:

1. `edit cluster-config -replication_port` always writes a `[replication_port://PORT]` (non-SSL) stanza to `server.conf`, overwriting the customer's `[replication_port-ssl://PORT]` stanza. PR #903 attempted to fix this by omitting `-replication_port` for SSL deployments, but Splunk's REST API **requires** the `-replication_port` parameter regardless of SSL mode. This caused the command to fail with `parameter=replication_port not present` and loop indefinitely — the indexer never joined the cluster.
2. Even after passing `-replication_port` and re-applying the SSL stanza, the plain `[replication_port://PORT]` stanza `edit cluster-config` writes is never removed. `splunkd` loads the non-SSL stanza first, so replication silently falls back to plaintext unless the stale stanza is explicitly deleted.
3. The task that re-applies the SSL replication stanza after cluster join originally routed through `set_config_file.yml`. On some published `splunk/splunk` base images (confirmed on `10.2.5`), that helper **deletes and fully rewrites `server.conf`** before writing the passed-in stanzas. Since only the SSL replication stanza is passed, this wipes out the `[clustering]` stanza and `site=` setting the cluster-join task had just written moments earlier — permanently unregistering the peer from the ClusterManager (the pod reports `1/1 Ready`, but the peer never appears in `splunk list cluster-peers`, or shows up with an empty ID).

## Fix

1. **Always pass `-replication_port`** so `edit cluster-config` succeeds (REST API requirement).
2. **Remove the stale non-SSL `[replication_port://PORT]` stanza** left behind by `edit cluster-config` so SSL replication doesn't silently fall back to plaintext.
3. **Re-apply the SSL replication stanza via `set_config_stanza.yml`** (a surgical `ini_file`/`lineinfile` upsert of just the targeted stanza) instead of `set_config_file.yml`, so `[clustering]`/`site=` are never touched.

SSL mode is detected via either:
- Explicit `splunk.idxc.replication_ssl: true` flag, **or**
- Auto-detection of `replication_port-ssl://` keys in `splunk.conf.server.content`

This means customers with only the conf-based SSL config (no extra flag) are handled automatically.

## Files Changed

- `roles/splunk_indexer/tasks/indexer_clustering.yml` — single-site cluster join
- `roles/splunk_indexer/tasks/setup_multisite.yml` — multi-site cluster join

## Verification

Tested on Splunk Enterprise **10.4.0** on EKS (k8s 1.34), and separately re-verified end-to-end against the customer's exact base image (`splunk/splunk:10.2.5`) on a 2-site indexer cluster (4 indexers, 1 ClusterManager, 1 LicenseManager) reproducing their SSL replication config.

**Before fix:** `parameter=replication_port not present` error on every retry (issue 1); or, once that's worked around, the indexer joins but the SSL stanza reverts to non-SSL, or the peer disappears from the ClusterManager entirely on certain base images (issues 2/3).

**After fix:** on first startup, no manual restart needed:
```
[clustering]
mode = peer
site = site1                       ← survives (previously wiped by set_config_file.yml on some images)

[replication_port-ssl://9887]      ← restored by re-apply task, via set_config_stanza.yml
serverCert = ...
```
No bare `[replication_port://9887]` stanza remains. `splunk list cluster-peers` on the ClusterManager shows the peer `status:Up` with `replication_use_ssl:1`.

## Backward Compatibility

- Non-SSL deployments: unchanged (re-apply/removal tasks are skipped when `use_ssl_replication` is false)
- SSL deployments: fixed — SSL stanza and cluster membership now both persist on first startup

Jira: https://splunk.atlassian.net/browse/CSPL-4006

### DIFF
--- a/roles/splunk_indexer/tasks/indexer_clustering.yml
+++ b/roles/splunk_indexer/tasks/indexer_clustering.yml
@@ -14,7 +14,10 @@
   set_fact:
     use_ssl_replication: "{{ (splunk.idxc.replication_ssl | default(false) | bool) or (has_ssl_replication_port_in_conf | default(false) | bool) }}"
 
-- name: Set current node as indexer cluster peer (non-SSL replication)
+# CSPL-4006: Always pass -replication_port (Splunk REST API requires it regardless of SSL mode).
+# For SSL deployments the -replication_port value gets overwritten by Splunk CLI into a non-SSL
+# stanza; the re-apply task below restores the SSL stanza immediately after.
+- name: Set current node as indexer cluster peer
   command: "{{ splunk.exec }} edit cluster-config -mode slave -master_uri '{{ cert_prefix }}://{{ splunk.cluster_master_url }}:{{ splunk.svc_port }}' -replication_port {{ splunk.idxc.replication_port }} -secret '{{ splunk.idxc.pass4SymmKey }}' -auth '{{ splunk.admin_user }}:{{ splunk.password }}'"
   become: yes
   become_user: "{{ splunk.user }}"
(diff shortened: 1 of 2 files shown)


mgimlewi:
Hi @gabrielm-splunk, I believe I'm experiencing this exact issue as my indexer pods attempt to join my cluster (running Splunk Enterprise 9.4.x, currently 9.4.10).

I'm unable to view the Jira link used for tracking - but thought I'd inquire about whether this fix will be merged / ported to a 9.4.x release / a potential timeline for that?

Thank you for the find here!

PR #928, opened 2026-09-03
Add SPLUNK_KVSTORE_DISABLED to opt out of the KV Store

tod-uma:
## Why

Splunk indexers and forwarders are documented as safe to run without the App Key Value Store ([About the app key value store](https://help.splunk.com/en/splunk-enterprise/administer/admin-manual/9.2/administer-the-app-key-value-store/about-the-app-key-value-store)), but the collection has no way to express that. The only existing route is the generic `splunk.conf` escape hatch, and the one branch that tried this directly (`CSPL-4011-disable-kvstore`) is ingestor-scoped, unmerged, and puts `kvstore:` directly under `value:` — omitting the required `content:` level, so it never actually writes anything.

## What

Mirrors the newest kvstore task's presence-gated pattern (`set_default_kvstore_type.yml`):

- `roles/splunk_common/tasks/set_kvstore_disabled.yml` (new) — single `ini_file` task writing `[kvstore] disabled` to `server.conf`, following the register-and-restart pattern used elsewhere in this role (`set_server_name.yml`, `set_postgres_sidecar.yml`).
- Wired into `roles/splunk_common/tasks/main.yml` right after `set_default_kvstore_type.yml`, gated on **presence** (`'disabled' in splunk.kvstore`) rather than a `false` default — same reasoning upstream already applied to `default_kvstore_type`, and it means `roles/splunk_indexer/molecule/default/tests/test_default.py`'s `mongod.lock` assertion is untouched (this is opt-in, not a default-behavior change).
- `SPLUNK_KVSTORE_DISABLED` env override in `inventory/environ.py`, mirroring `getDefaultKVStoreType`'s shape.
- Docs in `docs/ADVANCED.md` and `docs/advanced/default.yml.spec.md`.
- Unit test in `tests/small/test_environ.py`.

## Testing

- `tests/small/` (pytest): all new cases pass, no regressions.
- Functional dry-run: ran the actual `main.yml` gate condition and `set_kvstore_disabled.yml` task via `ansible-playbook` against a scratch `splunk.home`, confirmed the gate only fires when `kvstore.disabled` is explicitly set, the first write is `changed`, a repeat is idempotent, and the resulting `server.conf` reads exactly `[kvstore]\ndisabled = <bool>`.
- **Not yet run**: this hasn't been exercised on a live Splunk instance. We're building it as the KV-store-disable pilot for a new production indexer joining our cluster and expect to run it there within the next few days — will report back with the result.

### DIFF
--- a/docs/ADVANCED.md
+++ b/docs/ADVANCED.md
@@ -87,6 +87,7 @@ Splunk-Ansible ships with an inventory script in `inventory/environ.py`. The scr
 | SPLUNKD_SSL_PASSWORD | Custom SSL password used with Splunkd when HTTPS is enabled | no | no | no |
 | SPLUNK_KVSTORE_PORT | Port to run Splunk KVStore. Default: `8191` | no | no | no |
 | SPLUNK_KVSTORE_DEFAULT_TYPE | Configures `[kvstore] defaultKVStoreType` in `server.conf` when explicitly set. Allowed values are `cohosted` and `local`; Ansible writes the setting only for Splunk Enterprise 10.6.0 or later. When set to `cohosted`, Ansible also writes `[kvstore] postgresMigrateOnStartup=true`; when set to `local`, it writes `[kvstore] postgresMigrateOnStartup=false`. Older Splunk versions skip the default KV Store type configuration. SOK injects `local` for supported CMP-K deployments. | no | no | no |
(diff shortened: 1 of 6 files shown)


tod-uma:
Pushed a fix (`9a662cb`): the new `set_kvstore_disabled` task was missing `become`/`become_user`, which every other task in this file that writes into `$SPLUNK_HOME` sets explicitly (e.g. `set_default_kvstore_type.yml`, `set_general_symmkey_password.yml`) since this role's play-level `become` isn't assumed. Found building a real new indexer against this branch — invisible in any environment where the KV store is already disabled, since the `ini_file` write becomes a no-op there.

PR #929, opened 2026-09-03
Add explicit Noah role provisioning

patrykw-splunk:
## Summary

Add explicit Noah role provisioning to splunk-ansible with support for both dict and list `splunk.conf` formats.

- **Explicit Noah role (`splunk_noah`)**: New Ansible role activated only when `SPLUNK_NOAH_ENABLED=true`. Provides two entry points: `pre_auth.yml` (writes a disabled `[noahService]` stanza before temporary auth startup) and `post_config.yml` (applies role-specific profile before first full splunkd start). Supported roles: indexer (peer), search head (search client), deployer (none).
- **Dual-format `splunk.conf` support**: Noah tasks detect whether `splunk.conf` is a dict or list and extract the `[noahService]` stanza accordingly. The same dual-format pattern is applied to SSL replication port checks in `indexer_clustering.yml` and `setup_multisite.yml`.
- **Split server.conf normalization**: List-form `server.conf` entries supplied separately (e.g. ConfigMap + Secret) are recursively merged by effective file path before any role consumes them. This allows SOK to deliver structural `[noahService]` values through a ConfigMap and `pass4SymmKey` through a Secret independently.
- **Dedicated Noah `pass4SymmKey` path**: Removed the fallback from `splunk.pass4SymmKey` (the Splunk-to-Splunk `[general]` key) to prevent overwriting the correct Noah key staged from a Kubernetes Secret.
- **SHC restart deduplication**: Avoid duplicate restarts in both classic and Noah modes during search-head cluster formation.
- **Deployer pre-start configuration**: Configure SHC deployer before initial start so it is ready when search heads join.

#### Test plan

- [ ] Verify Noah indexer provisioning with list-form `splunk.conf` (ConfigMap + Secret split)
- [ ] Verify Noah search head provisioning with dict-form `splunk.conf`
- [ ] Confirm classic (non-Noah) mode is unaffected
- [ ] Run `tests/small/test_noah_role_configuration.py` and `tests/small/test_environ.py`


roles/splunk_common/tasks/configure_deployer_prestart.yml
```diff
@@ -0,0 +1,69 @@
+---
```
minjieqiu, line 1: The tasks in this file should not executed if the deployer url is empty.

roles/splunk_common/tasks/configure_shc_prestart.yml
```diff
@@ -0,0 +1,186 @@
...
+        and (splunk.set_search_peers | default(false) | bool)
+      }}
+
+- name: Validate pre-start SHC configuration inputs
+  assert:
+    that:
+      - cert_prefix in ["http", "https"]
+      - shc_prestart_server_name | length > 0
+      - splunk.hostname | default("") | length > 0
+      - splunk.deployer_url | default("") | length > 0
```
minjieqiu, line 22: splunk.deployer_url can be empty.

roles/splunk_common/tasks/configure_shc_prestart.yml
```diff
@@ -0,0 +1,186 @@
+---
+- name: Set expected pre-start SHC configuration
+  set_fact:
+    shc_prestart_server_name: "{{ splunk.server_name | default(splunk.hostname, true) }}"
+    shc_prestart_mgmt_uri: "{{ cert_prefix }}://{{ splunk.hostname }}:{{ splunk.svc_port }}"
+    shc_prestart_deployer_uri: "{{ cert_prefix }}://{{ splunk.deployer_url }}:{{ splunk.svc_port }}"
```
minjieqiu: splunk.deployer_url can be empty.

roles/splunk_common/tasks/configure_shc_prestart.yml
```diff
@@ -0,0 +1,186 @@
+---
+- name: Set expected pre-start SHC configuration
```
minjieqiu: For the SHC bootstrap optimization which I mostly are in this file, let’s wrap that logic in a conditional check so it only triggers when noah is enabled.

roles/splunk_common/tasks/configure_deployer_prestart.yml
patrykw-splunk: Please check last commit. I think it should fix it

roles/splunk_common/tasks/configure_shc_prestart.yml
patrykw-splunk: Please check last commit. I think it should fix it

jmeixensperger (dismissed):

ruomeiy-splunk (commented):
Inline review comments for the regressions identified in PR #929. Posting as comments only; this review does not change the approval state.

roles/splunk_indexer/tasks/indexer_clustering.yml
```diff
@@ -53,17 +76,9 @@
   vars:
     conf_file: "server.conf"
     conf_directory: "{{ splunk.home }}/etc/system/local"
-    stanza_name: "{{ stanza.key }}"
-    stanza_settings: "{{ stanza.value | default({}) }}"
-  with_dict: "{{ splunk.conf.server.content }}"
-  loop_control:
-    loop_var: stanza
+    conf_stanzas: "{{ idxc_server_conf_content }}"
```
ruomeiy-splunk: [P1] This includes `set_config_stanza.yml` but only supplies `conf_stanzas`. The included tasks require `stanza_name` and `stanza_settings`, so any detected SSL replication stanza reaches an undefined variable and aborts provisioning. Restore the per-stanza loop/variables or include `set_config_file.yml`.

roles/splunk_indexer/tasks/setup_multisite.yml
```diff
@@ -13,9 +13,32 @@
...
+      {{ ((splunk_conf_effective | default(splunk.conf))
+          | selectattr('key', 'equalto', 'server') | first | default({}))
+          .get('value', {}).get('content', {}) }}
+  when:
+    - "'conf' in splunk and splunk.conf"
+    - splunk.conf is not mapping
+
+- name: Check if SSL replication port is configured in server.conf (multisite)
+  set_fact:
+    has_ssl_replication_port_in_conf_multisite: >-
```
ruomeiy-splunk: [P1] This task sets `has_ssl_replication_port_in_conf_multisite`, but the `Determine SSL replication mode` task below still reads `has_ssl_replication_port_in_conf`. In a multisite-only configuration where the single-site task did not populate that old fact, an SSL stanza with `replication_ssl=false` is treated as plaintext and the non-SSL listener is retained.

roles/splunk_indexer/tasks/indexer_clustering.yml
```diff
@@ -10,9 +10,32 @@
...
+  when:
+    - "'conf' in splunk and splunk.conf"
+    - splunk.conf is mapping
+    - "'server' in splunk.conf"
+    - "'content' in splunk.conf.server"
+
+- name: Extract server.conf content from splunk.conf (list format)
+  set_fact:
+    idxc_server_conf_content: >-
+      {{ ((splunk_conf_effective | default(splunk.conf))
```
ruomeiy-splunk: [P1] The new list-form SSL detection happens only after `configure_indexer_cluster_prestart.yml`, whose SSL check handles mapping-form configuration only. With list-form SSL and `replication_ssl=false`, prestart creates a plaintext `replication_port://` stanza and marks the node preseeded; the later cleanup is then skipped. This selection is also order-dependent because it takes the first `server` entry without matching its effective directory. Normalize/select the effective system-local configuration before pre-seeding and reuse it here.

inventory/environ.py
```diff
@@ -194,7 +197,22 @@ def getServiceName(vars_scope):
     if serviceName != "" and namespace != "":
         vars_scope["splunk"]["issuer_uri"] = "{}.{}.svc.{}".format(serviceName, namespace, clusterDomain)
     if headlessServiceName != "" and namespace != "":
-        vars_scope["splunk"]["server_name"] = "{}.{}.{}.svc.{}".format(podName, headlessServiceName, namespace, clusterDomain)
+        server_name = "{}.{}.{}.svc.{}".format(podName, headlessServiceName, namespace, clusterDomain)
+        vars_scope["splunk"]["server_name"] = server_name
+        if vars_scope.get("splunk_noah_enabled", False):
+            vars_scope["splunk"]["noah_advertised_addr"] = "https://{}:{}".format(server_name, vars_scope["splunk"]["svc_port"])
```
ruomeiy-splunk: [P1] Setting `noah_advertised_addr` here makes the later resolver mistake this generated HTTPS value for a defaults-file override. It overwrites an explicitly configured address, advertises HTTPS when splunkd SSL is disabled, and bypasses the resolver's required `POD_NAME` validation. Leave advertised-address resolution to `getNoahAdvertisedAddr()`.

roles/splunk_common/tasks/main.yml
```diff
@@ -134,10 +140,41 @@
...
+    - splunk.role == "splunk_search_head" or splunk.role == "splunk_search_head_captain"
+    - splunk_search_head_cluster | bool
+    - splunk_status.rc != 0
+
+- name: Configure SHC deployer before splunkd starts
+  include_tasks: configure_deployer_prestart.yml
+  when:
+    - ansible_system is match("Linux")
+    - splunk.role == "splunk_deployer"
+    - splunk.deployer_url | default("") | length > 0
```
ruomeiy-splunk: [P1] The shipped `deployer_url` default is `null`, and `default("")` does not replace a defined `None`. Evaluating `length` therefore raises `object of type 'NoneType' has no len()` for any Linux deployer without `SPLUNK_DEPLOYER_URL`, before the later conditions can skip this include. Use `default("", true)` at minimum, or remove the deployer's self-URL requirement.

inventory/environ.py
```diff
@@ -1153,32 +1159,41 @@ def loadHostDefaults(config):
...
+
+def redact_sensitive_keys(node, stars):
+    """
+    Recursively replace the value of any sensitive key with stars. Walking the whole
+    tree covers arbitrary user-supplied structures such as splunk.conf stanzas, in
+    both the dictionary and list forms accepted by the configuration writer.
+    """
+    if isinstance(node, dict):
+        for key, value in node.items():
+            if isinstance(value, (dict, list)):
```
ruomeiy-splunk: [P1] Sensitive parent keys are checked only after dictionary/list values are recursively traversed. Consequently values such as `password: [LEAKME]` or `secret: {raw: LEAKME}` remain plaintext; `develop` redacts those whole values. Check whether the key is sensitive first, regardless of the value type.

site.yml
```diff
@@ -39,7 +39,9 @@
 
         - name: Check all instances for required restarts
           include_tasks: ./roles/splunk_common/tasks/check_for_required_restarts.yml
-          when: splunk_restart_triggered is not defined or not splunk_restart_triggered
+          when:
+            - splunk_restart_triggered is not defined or not splunk_restart_triggered
+            - not (shc_prestart_defer_initial_restart | default(false) | bool)
```
ruomeiy-splunk: [P2] `shc_prestart_defer_initial_restart` remains true through role work and `ansible_post_tasks`, so this suppresses the final global restart-required probe. A later setting change that does not explicitly notify a handler can finish successfully while remaining unapplied. Scope or clear the bootstrap deferral before this final check.

roles/splunk_common/tasks/configure_shc_prestart.yml
```diff
@@ -0,0 +1,201 @@
...
+    path: "{{ shc_prestart_server_conf }}"
+    section: "shclustering"
+    option: "conf_deploy_fetch_url"
+    value: "{{ shc_prestart_deployer_uri }}"
+    allow_no_value: false
+    state: present
+    mode: 0660
+    owner: "{{ splunk.user }}"
+    group: "{{ splunk.group }}"
+  when: shc_prestart_deployer_uri | length > 0
```
ruomeiy-splunk: [P2] When the desired deployer URI is empty, skipping this task does not remove an existing `conf_deploy_fetch_url`. A persistent SHC migrated from deployer-backed configuration to deployerless Noah therefore retains the stale endpoint, while the validation below accepts the empty desired URI and later initialization is skipped. Reconcile the option to `absent` when the desired URI is empty.

ruomeiy-splunk (dismissed):

roles/splunk_indexer/tasks/setup_multisite.yml
patrykw-splunk: Can you check if now it looks correct? thanks 🙇 
I adressed it in 29f245abefc4236aafdb97ea6aedf2c128031856

roles/splunk_indexer/tasks/indexer_clustering.yml
patrykw-splunk: Can you check if now it looks correct? thanks 🙇 
I adressed it in 29f245abefc4236aafdb97ea6aedf2c128031856

roles/splunk_indexer/tasks/indexer_clustering.yml
patrykw-splunk: I adressed this issue here 9a4e053f0a1bdbdd03dde834d8a954ba90de3605

inventory/environ.py
patrykw-splunk: I adrdessed this issue here - cc1d88e79e4947af051e45bfaf4811e9adb4ba4a

roles/splunk_common/tasks/main.yml
patrykw-splunk: I adressed this issue here - 02537f23446cbabd9ad153b43c3dfd234b4443f0

inventory/environ.py
patrykw-splunk: I addressed this issue here - 48ed8c82624715f51bca7828029e5fefd335ef72

site.yml
patrykw-splunk: I adressed this issue here - 02e85dbf7e337e023d1496cdabe4b97b671a1a59

roles/splunk_common/tasks/configure_shc_prestart.yml
patrykw-splunk: I adressed this issue here - https://github.com/splunk/splunk-ansible/pull/929/changes/7fa22d98e0c4c46e90f5c72af404c8196adae2bf

site.yml
```diff
@@ -37,11 +37,18 @@
             - ansible_post_tasks_item is not none
             - ansible_post_tasks_item is match("^(http|https|file)://.*")
 
+        # Clear the SHC bootstrap deferral so the final restart probe is never
+        # suppressed. The flag is only meant to avoid redundant restarts during
+        # the search-head role work, not to skip this site-level safety net.
+        - name: Clear SHC pre-start restart deferral before final check
+          set_fact:
+            shc_prestart_defer_initial_restart: false
```
jmeixensperger: Is this doing anything with the latest changes? The check for required restarts task below no longer uses the var, so all uses should be contained in the `Provision role` task above. In fact, I only see the var being used in `configure_shc_prestart.yml`, which is a part of the role provisioning.

ruomeiy-splunk (approved):

PR #940, opened 2026-09-17
fix: bootstrap SHC without requiring a deployer

minjieqiu:
## Summary

`splunk_search_head_cluster` (`roles/splunk_common/tasks/get_facts.yml`, derived purely from `search_head_captain_url`) already identifies which search-head pods need SHC bootstrap, independent of whether a deployer exists. The gate in `main.yml` additionally required `deployer_url` to be set, so a deployerless SHC topology (e.g. Noah, which has no deployer) never bootstrapped clustering at all and stayed permanently unclustered.

## Changes

- `roles/splunk_search_head/tasks/main.yml`: drop the `deployer_url` requirement from the `search_head_clustering.yml` include gate — `splunk_search_head_cluster | bool` alone is sufficient.
- `roles/splunk_search_head/tasks/search_head_clustering.yml`:
  - The two `wait_for_splunk_instance` checks against `splunk.deployer_url` now skip cleanly when it isn't set, instead of waiting on an address that doesn't exist.
  - The `-conf_deploy_fetch_url` argument to `init shcluster-config` is now an optional Jinja-computed fragment, omitted when there's no deployer, instead of always being baked in.

Confirmed via a repo-wide search that these two files are the only places `deployer_url`/`conf_deploy_fetch_url` affect SHC bootstrap behavior. Everything else — captaincy determination, `bootstrap shcluster-captain`, `add shcluster-member`, `resync shcluster-replicated-config` — already depends only on `search_head_captain_url`, never `deployer_url`, so none of it needed to change.

## Test plan

- [x] YAML syntax validated for both modified files
- [x] Jinja expression for the optional `-conf_deploy_fetch_url` fragment rendered standalone for all three cases: deployer set, empty string, key missing entirely
- [ ] Live SHC bootstrap test on a deployerless (Noah) topology
- [ ] Regression check that classic SHC-with-deployer bootstrap is unaffected


ruomeiy-splunk:
Please update `docs/ADVANCED.md`, in the **Supported environment variables** table, on the `SPLUNK_DEPLOYER_URL` row (currently around line 38). Change the **Required for Search Head Clustering** value from `yes` to `no`.

Consider also updating the description to clarify: `Optional; required only when SHC deployer functionality is used.`

This keeps the documentation consistent with this MR’s new deployerless SHC bootstrap support.

ruomeiy-splunk (approved):

jmeixensperger (approved):

